PR #7, opened 2020-06-22
Add i/o heavy insight

uhoh-itsmaciek:


explain/insights/io-heavy.md
```diff
@@ -0,0 +1,17 @@
...
+explain_insight: 'i/o-heavy'
+title: 'EXPLAIN - Insights: I/O Heavy'
+backlink_href: /docs/explain/insights
+backlink_title: 'Documentation: EXPLAIN - Insights'
+---
+
+**Description:**
+
+Postgres will often spend most of its I/O time on a query in a small set of plan nodes. This insight identifies the most expensive nodes (according to time spent doing I/O) to guide optimization. Since I/O is usually the bottleneck in many database queries, focusing on nodes performing
+the bulk of plan I/O is a good rule of thumb.
```
lfittl, line 11: I think “plan I/O” is a bit odd here - maybe just “I/O”

explain/insights/io-heavy.md
```diff
@@ -0,0 +1,17 @@
...
+
+This insight requires `track_io_timing` to be enabled on the database.
+
+**Recommended Action:**
+
+If possible, reduce the amount of I/O performed in this node. There are many possible causes of excessive I/O, but several are identified by  other insights, like [Inefficient Index](/docs/explain/insights/inefficient-index) or [Large Offset](/docs/explain/insights/large-offset).
```
lfittl, line 17: Accidental double space after “identified by”

explain/insights/io-heavy.md
```diff
@@ -0,0 +1,17 @@
+---
+explain_insight: 'i/o-heavy'
```
lfittl, line 2: Is the “/“ in the insight name intentional? (i.e. isn’t it “io-heavy” without the slash)

explain/insights/io-heavy.md
uhoh-itsmaciek: Yeah, I think you're right. It read a little funny to me just bare so I added in a filler word, but it does work without it.

explain/insights/io-heavy.md
uhoh-itsmaciek: The annotation "slug" in pganalyze is "i/o-heavy" right now. I felt that read better than the alternatives. That's what this is referring to, no?

PR #8, opened 2020-06-23
Reference Amazon Linux 2 instead of original

uhoh-itsmaciek:
See https://github.com/pganalyze/pganalyze/issues/744


uhoh-itsmaciek:
See also https://github.com/pganalyze/pganalyze/pull/755

PR #19, opened 2020-09-10
Improve collector documentation

uhoh-itsmaciek:
This is not ready to merge, but wanted to put it out there for review. I documented the immediately obvious settings, and arranged things according to how they are in `config/config.go`. This means that the general settings / additional settings need some re-arranging (e.g., `ignore_schema_regexp` is probably more useful to highlight than `api_base_url`), but I feel like having both sections makes sense since some settings are way less common.

I added question marks where I was not sure about defaults or the description (sometimes there is only a question mark). Some of these I can probably figure out on my own; but I wanted to get the full structure out there first.

I'll add a pganalyze-web PR where we can have a review app.


collector/index.md
```diff
@@ -0,0 +1,425 @@
+---
+title: 'pganalyze Collector'
+backlink_href: /docs
+backlink_title: 'Documentation'
+---
+
+The pganalyze collector periodically queries your database and sends metrics and metadata (as "snapshots") to
+the pganalyze app. This is the only component that talks directly to your database. For details on setting up
```
lfittl, line 8: "This is the only component that talks directly to your database" could be emphasized more - maybe we can create a dedicated security section?

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+the collector, see the [collector installation guide](/docs/install/02_installing_the_collector). The collector
+is open-source and the code is available [on GitHub](github.com/pganalyze/collector).
```
lfittl, line 10: this link is missing a leading https://

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+
+
+# Configuration Settings
+
+The collector can be configured either through a `INI` config file (in a normal install) or environment variables
```
lfittl, line 15: `a` => `an` INI file

lfittl: "normal install" => "package-based install"

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+(typically for running via its Docker image). Most settings can be configured through either mechanism (if both
+are present, the config file takes precedence). Note that a single collector instance can monitor more than one
+database server (though this is not supported when configuring through environment variables). The `[pganalyze]`
```
lfittl, line 18: (though this is not supported when configuring through environment variables)

=> `This is not supported when configuring through environment variables.`

lfittl: And make a new paragraph after this line.

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+section describes settings that apply to all servers, and the other sections are named after the database server
+they correspond to, and define settings for just that server. When using environment variables, the setting applies
```
lfittl, line 20: `sections are named after the database server they correspond to` => we should clarify that its a name you chose (doesn't have to correlate with the server name, or hostname - in fact the example we usually give is `server1` and `server2`)

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+to all monitored servers.
+
+After you make changes, the config file can be reloaded while the collector is running with the command
+`pganalyze-collector --reload`. This minimizes monitoring interruptions and simplifies config file updates.
```
lfittl, line 24: "After you make changes, you can run `pganalyze-collector --test --reload` to verify the new configuration and reload the collector if they work correctly. This minimizes monitoring interruptions and simplifies config file updates."

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+## General settings
+
+Common settings for configuring monitoring regardless of platform.
+
+<table>
+  <thead>
+    <tr>
+      <th>setting</th>
+      <th>default</th>
+      <th>description</th>
```
lfittl, line 40: This feels better with uppercase leading letters.

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>
+        API key to authenticate the collector to the pganalyze app. We will show this
+        when adding a server in the app, and you can review it in your organization's
+        API keys page.
+      </td>
+    </tr>
+    <tr>
+      <td>api_base_url (PGA_API_BASEURL)</td>
+      <td>https://api.pganalyze.com</td>
+      <td>URL to contact the pganalyze app. You typically will not need to change this.</td>
```
lfittl, line 56: Base URL for contacting the pganalyze API. You typically do not need to change this, unless you are running [pganalyze Enterprise Edition](/docs/enterprise).

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+    <tr>
+      <td>error_callback</td>
```
lfittl, line 59: Move to additional settings

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>[none]</td>
+      <td>
+        Script to call if snapshot fails (learn more <a target="_blank" href="https://github.com/pganalyze/collector/#successerror-callbacks">in the collector README</a>)
+      </td>
+    </tr>
+    <tr>
+      <td>success_callback</td>
```
lfittl, line 66: Move to additional settings

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>[none]</td>
+      <td>
+        Script to call if snapshot succeeds (learn more <a target="_blank" href="https://github.com/pganalyze/collector/#successerror-callbacks">in the collector README</a>)
+      </td>
+    </tr>
+    <tr>
+      <td>enable_reports (PGA_ENABLE_REPORTS)</td>
+      <td>false</td>
+      <td></td>
```
lfittl, line 75: We can just skip this (or comment it out)

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+    <tr>
+      <td>disable_logs (PGA_DISABLE_LOGS)</td>
```
lfittl, line 78: move to additional settings

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>false</td>
+      <td>Disable log collection and processing</td>
+    </tr>
+    <tr>
+      <td>disable_activity (PGA_DISABLE_ACTIVITY)</td>
```
lfittl, line 83: move to additional settings

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>false</td>
+      <td>Disable activity snapshots and processing (current vacuums, queries, etc.)</td>
+    </tr>
+    <tr>
+      <td>enable_log_explain (PGA_ENABLE_LOG_EXPLAIN)</td>
```
lfittl, line 88: this is good to keep in the beginning (as it is now)

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>db_sslkey_contents (DB_SSLKEY_CONTENTS)</td>
+      <td>[none]</td>
+      <td>as above, but actual contents of the key</td>
+    </tr>
+  </tbody>
+</table>
+
+## AWS settings
+
+Only relevant if you are running your database in AWS RDS. Note that the <code>aws_endpoint_*</code>
```
lfittl, line 183: If you use the built-in markdown tags this gets escaped correctly (backticks), otherwise you need to prefix underscores like this: `\_`

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+  <tbody>
+    <tr>
+      <td>db_url (DB_URL)</td>
+      <td>n/a - either this or individual db_* settings are required</td>
+      <td>URL of the database server to monitor</td>
+    </tr>
+    <tr>
+      <td>db_name (DB_NAME)</td>
+      <td>n/a - either this or db_url is required</td>
+      <td>name of database to monitor; or, comma-separated list of all databases to monitor,
```
lfittl, line 120: I think it would be good to uppercase the first letter of the description

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>azure_db_server_name (AZURE_DB_SERVER_NAME)</td>
+      <td>n/a, required if azure?</td>
+      <td>?</td>
+    </tr>
+    <tr>
+      <td>azure_eventhub_namespace (AZURE_EVENTHUB_NAMESPACE)</td>
```
lfittl, line 263: We could reference the Log Insights documentation here (which explains how to use this setting)

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    <tr>
+      <th>setting</th>
+      <th>default</th>
+      <th>description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>azure_db_server_name (AZURE_DB_SERVER_NAME)</td>
+      <td>n/a, required if azure?</td>
```
lfittl, line 259: Auto-detected from hostname

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+  </thead>
+  <tbody>
+    <tr>
+      <td>aws_region (AWS_REGION)</td>
+      <td>n/a, required if aws</td>
+      <td>region your AWS server is running in</td>
+    </tr>
+    <tr>
+      <td>aws_db_instance_id (AWS_INSTANCE_ID)</td>
+      <td>n/a, required if aws</td>
```
lfittl, line 203: Auto-detected from hostname

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    <tr>
+      <th>setting</th>
+      <th>default</th>
+      <th>description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>aws_region (AWS_REGION)</td>
+      <td>n/a, required if aws</td>
```
lfittl, line 198: Auto-detected from hostname

collector/index.md
lfittl: May need to be set manually when using IP addresses, or custom DNS records.

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>region your AWS server is running in</td>
+    </tr>
+    <tr>
+      <td>aws_db_instance_id (AWS_INSTANCE_ID)</td>
+      <td>n/a, required if aws</td>
+      <td>the instance id of your AWS server</td>
+    </tr>
+    <tr>
+      <td>aws_access_key_id (AWS_ACCESS_KEY_ID)</td>
+      <td>n/a, required if aws and using logs?</td>
```
lfittl, line 208: This is not needed when using instance roles (which we can get from the metadata service when using an EC2 instance)

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    <tr>
+      <th>setting</th>
+      <th>default</th>
+      <th>description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>gcp_cloudsql_instance_id (GCP_CLOUDSQL_INSTANCE_ID)</td>
+      <td>n/a, required if gcp?</td>
```
lfittl, line 315: This may be auto-detected from hostname as well (not sure), double check the code.

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+      <td>?</td>
+    </tr>
+    <tr>
+      <td>gcp_pubsub_subscription (GCP_PUBSUB_SUBSCRIPTION)</td>
+      <td>n/a, required if gcp?</td>
```
lfittl, line 320: Reference log insights docs here

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+    <tr>
+      <td>api_system_scope (PGA_API_SYSTEM_SCOPE)</td>
+      <td>?</td>
+      <td>?</td>
+    </tr>
+    <tr>
+      <td>db_log_location (LOG_LOCATION)</td>
+      <td>?</td>
+      <td>database log file or directory location (must be readable by pganalyze system user)</td>
```
lfittl, line 367: This here could be moved to a dedicated section for self-managed servers.

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+    <tr>
+      <td>db_log_docker_tail</td>
+      <td>[none]</td>
+      <td>experimental: name of docker container to collect logs from using <code>docker logs -t</code></td>
```
lfittl, line 372: This can be documented out for now (we just use it for development purposes)

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+    <tr>
+      <td>ignore_schema_regexp (IGNORE_SCHEMA_REGEXP)</td>
+      <td>[none]</td>
+      <td>do not monitor matching tables, schemas, or functions; match is checked against schema-qualified object names (e.g.
+      to ignore table "foo" only in the public schema, set to <code>^public\.foo$</code>)
+      </td>
```
lfittl, line 379: We could move this up into the general section

collector/index.md
```diff
@@ -0,0 +1,425 @@
...
+    </tr>
+    <tr>
+      <td>filter_query_sample (FILTER_QUERY_SAMPLE)</td>
+      <td>none</td>
+      <td>none/all</td>
+    </tr>
+    <tr>
+      <td>filter_query_text (FILTER_QUERY_TEXT)</td>
+      <td>unparsable</td>
+      <td>none/unparsable</td>
```
lfittl, line 407: These settings would be best in a dedicated "PII Filtering" section

collector/index.md
uhoh-itsmaciek: Does not look like it, but then I don't see where this happens for AWS either. I searched for [AwsDbInstanceID](https://github.com/pganalyze/collector/search?q=AwsDbInstanceID&unscoped_q=AwsDbInstanceID) and none of those look relevant. :confused:

lfittl: AWS and Azure happen here: 

https://github.com/pganalyze/collector/blob/8bd5b1b56df9423b10bdf284d636fe6ac5e153e7/config/read.go#L166

GCP uses IP addresses instead of hostnames, so the logic doesn't apply there - thus this is required when using Log Insights (and in general recommended, since uniquely identifies the server)

collector/index.md
uhoh-itsmaciek: Ended up just dropping the parentheses in the first comment, and adding a comma--I think that reads better and makes it clearer what the "this" is.

uhoh-itsmaciek:
@lfittl this probably still needs a few tweaks, and I had some questions in the Additional Settings section, but it could use another look.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+      <td>10</td>
+      <td>
+        Maximum connections allowed to the database with the collector application_name, in order to protect
+        against accidental connection leaks in the collector
+      </td>
+    </tr>
+    <tr>
+      <td>http_proxy (<code>HTTP_PROXY</code>)</td>
+      <td>[none]</td>
+      <td>?</td>
```
lfittl, line 461: Proxy to be used for all HTTP connections, such as API calls. Use for proxies that do not support SSL. Example: `http://username:password@myproxy`

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+    </tr>
+    <tr>
+      <td>https_proxy (<code>HTTPS_PROXY</code>)</td>
+      <td>[none]</td>
+      <td>?</td>
```
lfittl, line 466: Proxy to be used for all HTTP connections, such as API calls. Use for proxies that support SSL. Example: `https://username:password@myproxy`

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+    </tr>
+    <tr>
+      <td>no_proxy (<code>NO_PROXY</code>)</td>
+      <td>[none]</td>
+      <td>?</td>
```
lfittl, line 471: Comma-delimited list of hostnames that should be accessed directly, without using a configured proxy. Has no effect unless either `HTTP_PROXY` or `HTTPS_PROXY` is specified.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+    </tr>
+    <tr>
+      <td>disable_logs (<code>PGA_DISABLE_LOGS</code>)</td>
+      <td>false</td>
+      <td>Disable log collection and processing</td>
```
lfittl, line 476: Maybe `Disable Log Insights data collection`

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+    </tr>
+    <tr>
+      <td>disable_activity (<code>PGA_DISABLE_ACTIVITY</code>)</td>
+      <td>false</td>
+      <td>Disable activity snapshots and processing (current vacuums, queries, etc.)</td>
```
lfittl, line 481: `Disable activity snapshot data collection (VACUUM, connection traces, etc.)`

index.md
```diff
@@ -82,6 +82,8 @@ title: 'Documentation'
       * [Unique](/docs/explain/other-nodes/unique)
       * [Window Aggregate](/docs/explain/other-nodes/window-aggregate)
 
+* **[Collector](/docs/collector)**
```
lfittl, line 85: Maybe it would be good if we put the settings into their own page, so the `/docs/collector` page only has the introduction notes + security notes, and then links to the configuration settings page.

In the future we should move the monitoring user instructions from the collector README into its own docs page here, and it would make sense if that we have a "Collector overview" page then.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+
+Keeping your data safe is very important to us. We take a number of precautions to minimize the data
+we have access to, and to limit our level of access to that data. The collector is the only pganalyze
+component that talks directly to your database. In our recommended configuration, the collector uses
+a database role with limited access, and can only read system statistics and metadata&mdash;it cannot actually
+query your data.
+
+The collector connects to your database on the port you specify (default 5432) using the standard Postgres
+protocol, to Amazon S3 also over https (port 443) to upload monitoring data, and to the pganalyze API also
+over https. The collector only makes outbound connections&mdash;it does not listen on any ports.
```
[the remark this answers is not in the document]
uhoh-itsmaciek: That's great, thanks. I'm rewording the Heroku section slightly because the collector app is acting as the drain in this case (the drain in Heroku terminology is what consumes the logs, not the mechanism to get them there). But I think it's great how explicit we can be here.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+one database server, though this is not supported when configuring through environment variables.
+
+The `[pganalyze]` section describes settings that apply to all servers. Other sections describe the servers to
+monitor, how to connect to them, and server-specific configuration settings. You should name the other sections after
+the servers they correspond to, though note these are not the names that will appear in the app. In-app names are
+based on hostname settings as determined during monitoring. When using environment variables, the setting applies
+to all monitored servers.
+
+After you make changes, you can run `pganalyze-collector --test --reload` to verify the new configuration and load
+the new configuration in the collector service if they work correctly. This minimizes monitoring interruptions and
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Yeah, I was going for systemd/upstart terminology here, though I guess that might be confusing in this context. I feel like "background process" is vague but maybe that's fine; I'll change to that to at least make clear it's something local.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+      <td>[none]</td>
+      <td>Skip matching tables, schemas, or functions; match is checked against schema-qualified object names (e.g.
+      to ignore table "foo" only in the public schema, set to <code>^public\.foo$</code>)
+      </td>
+    </tr>
+    <tr>
+      <td>enable_log_explain (<code>PGA_ENABLE_LOG_EXPLAIN</code>)</td>
+      <td>false</td>
+      <td>Enable log-based EXPLAIN. See <a href="/docs/explain/setup/log_explain">setup instructions</a>, but note 
+      we recommend using <a href="/docs/explain/setup/auto_explain">autoexplain</a> instead if possible.
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Though just `auto_explain` since this is in embedded HTML, right?

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+    </tr>
+    <tr>
+      <td>db_username (<code>DB_USERNAME</code>)</td>
+      <td>n/a, either this or db_url is required</td>
+      <td>User to connect as</td>
+    </tr>
+    <tr>
+      <td>db_password (<code>DB_PASSWORD</code>)</td>
+      <td>n/a, either this or db_url is required</td>
+      <td>Password to connect with</td>
```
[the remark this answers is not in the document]
uhoh-itsmaciek: I wonder if this will confuse people with "Postgres user" vs. "`postgres` user"? But nah, it's probably fine. I'll go with this.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+      starting with primary (last entry can be <code>*</code> to monitor all databases on server)</td>
+    </tr>
+    <tr>
+      <td>db_username (<code>DB_USERNAME</code>)</td>
+      <td>n/a, either this or db_url is required</td>
+      <td>User to connect as</td>
+    </tr>
+    <tr>
+      <td>db_password (<code>DB_PASSWORD</code>)</td>
+      <td>n/a, either this or db_url is required</td>
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Or maybe just `[none]` for consistency with other places? I think if you're using the URL you probably know how to specify the password there.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+    </tr>
+    <tr>
+      <td>aws_assume_role (<code>AWS_ASSUME_ROLE</code>)</td>
+      <td>[none]</td>
+      <td>If using cross-account role delegation, the ARN of the role to assume; see the <a target="_blank" href="https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html">AWS documentation</a> for details</td>
+    </tr>
+    <tr>
+      <td>aws_endpoint_signing_region (<code>AWS_ENDPOINT_SIGNING_REGION</code>)</td>
+      <td>[none]</td>
+      <td>Region to use for signing requests</td>
```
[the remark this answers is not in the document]
uhoh-itsmaciek: This is very similar to 

```
<td>Path to the client SSL certificate (not needed for most configurations)</td>
```

above, but I like this wording better. I think we should be consistent; I'll make the change here and update that to match.

collector/index.md
```diff
@@ -0,0 +1,498 @@
...
+      <td>aws_endpoint_cloudwatch_logs_url (<code>AWS_ENDPOINT_CLOUDWATCH_LOGS_URL</code>)</td>
+      <td>[none]</td>
+      <td>URL of CloudWatch log service</td>
+    </tr>
+  </tbody>
+</table>
+
+## Azure settings
+
+Only relevant if you are running your database in Azure. See our <a href="https://pganalyze.com/docs/log-insights/setup/azure-database#step-2-setup-azure-event-hub">Azure setup instructions</a> for details.
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Good catch--this was initially inline in the table and I hoisted it for clarity. I'll drop the fragment.

uhoh-itsmaciek:
@lfittl thanks for the thorough feedback. I addressed all your comments and split the settings into their own page. I also adjusted the heading sizes to patch the other pages (I noticed we don't use `#` because that's the same size as the frontmatter `title`), and moved the settings title down a bit since most pages seem to have a brief intro to avoid two headings in a row.

collector/settings.md
```diff
@@ -0,0 +1,499 @@
...
+in the app. In-app names are based on hostname settings as determined during monitoring. When you set environment
+variables in addition to specifying a configuration file, the environment variable settings apply to all monitored
+servers.
+
+After you make changes, you can run `pganalyze-collector --test --reload` to verify the new configuration and load
+the new configuration in the collector background process if they work correctly. This minimizes monitoring
+interruptions and simplifies config file updates.
+
+
+## Configuration settings
```
lfittl, line 24: This heading also shows up in the TOC, which looks a bit odd. I wonder if we can just skip the heading, and make it one large intro section?

lfittl (approved):
Perfect - found one more thing, but otherwise lets 🚢

collector/settings.md
uhoh-itsmaciek: Yeah, you're right--I think that works better.

PR #28, opened 2020-12-31
Merge setup docs

uhoh-itsmaciek:
This merges the in-app documentation into this repo and converts all docs to MDX to better facilitate sharing.


components/SmartAnchor.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return <a {...props} />
+  }
+
+  const Link = linkComponent
+  destination = adjustLinks && destination && !destination.startsWith('/') ? '../' + destination : destination;
+  return <Link {...rest} to={destination} />
+};
+
+export function makeSmartAnchor(linkComponent: React.ComponentType<LinkProps>, linkRelative: boolean): React.ComponentType<Props> {
```
lfittl, line 54: As discussed, could be helpful to explain `linkRelative` logic.

components/SmartAnchor.tsx
```diff
@@ -0,0 +1,56 @@
...
+  let destination: string;
+  if ('to' in rest) {
+    destination = rest.to;
+    delete rest.to
+  } else {
+    destination = rest.href;
+    delete rest.href
+  }
+  const isFragmentLink = destination?.startsWith('#')
+  const isExternal = /^[a-zA-Z]+:/.test(destination)
```
lfittl, line 27: Maybe `isAbsolute` or `isFullURL` might be easier to understand

components/SmartAnchor.tsx
```diff
@@ -0,0 +1,56 @@
...
+  if (isBareAnchor) {
+    const props = {
+      ...rest,
+      href: destination,
+    }
+
+    if (isExternal) {
+      const rel = []
+      rel.push("noopener")
+      const isPganalyze = /^https?:\/\/[a-zA-Z-]+\.pganalyze\.com/.test(destination)
```
lfittl, line 39: `isPganalyzeWebLink` and then test for the `pganalyze.com` hostname only.

explain/index.mdx
```diff
@@ -51,5 +51,5 @@ pganalyze automatically analyzes EXPLAIN plans to find the most important insigh
 * **[Stale Stats](/docs/explain/insights/stale-stats)**<br />The table referenced has not had an `ANALYZE` run recently (potentially leading to inefficient plans)
 
 <br />
-<div><feature-page-cta event-label="Docs" title="Automatic Insights into Postgres Query Plans" description="pganalyze automatically gathers the auto_explain output on your behalf from the Postgres logs, so you can easily view the plans for slow queries without extra effort." path="/postgres-explain"></feature-page-cta></div>
+<div><FeaturePageCTA eventLabel="Docs" title="Automatic Insights into Postgres Query Plans" description="pganalyze automatically gathers the auto_explain output on your behalf from the Postgres logs, so you can easily view the plans for slow queries without extra effort." path="/postgres-explain" /></div>
 <br />
```
lfittl, line 55: It looks like the EXPLAIN index page broke, possibly due to the image changes requiring `visualization.png` to be included via an import instead of inline? (I don't think the CTA change is causing it)

Screen Shot 2021-01-16 at 2 25 41 PM

install/heroku_postgres/01_deploy_the_collector.mdx
```diff
@@ -0,0 +1,52 @@
...
+export const CollectorApp = ({ organizationSlug }) => (
+  <code>{collectorAppName(organizationSlug)}</code>
+);
+
+In order to monitor a Heroku Postgres database you need to run the pganalyze
+collector as an app inside your Heroku account. Through Heroku's add-on attachment system you then attach all databases that should be monitored.
+
+## Deploy collector
+
+First, deploy the collector to your own Heroku account with your pganalyze API key <APIKey apiKey={props.apiKey} />:
```
lfittl, line 22: We should probably have separate logic here for in-app (where we have an API key, vs public docs):

Screen Shot 2021-01-16 at 2 35 16 PM

(note the small empty `<code>` block)

install/heroku_postgres/01_deploy_the_collector.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+<p>
+  <a
+    href={`https://heroku.com/deploy?template=https://github.com/pganalyze/collector&env[PGA_API_KEY]=${props.apiKey}`}
```
lfittl, line 26: It looks like this link somehow doesn't work, since it gets prefixed with the docs page URL:

```
http://localhost:8000/docs/install/heroku_postgres/%7B%60https://heroku.com/deploy?template=https://github.com/pganalyze/collector&env[PGA_API_KEY]=${props.apiKey}`}
```

(this is the actual URL it tried to direct me to)

install/troubleshooting/rds_pg_stat_statements_shared_preload_libraries.md
```diff
@@ -1,21 +0,0 @@
----
-title: 'Amazon RDS: Resolving the "pg_stat_statements must be loaded via shared_preload_libraries" error'
```
lfittl, line 2: As mentioned in Slack, I think we should keep this page for now (and may want to make it more generic in the future, so it doesn't just talk about RDS, but rather in general addresses this setup error)

install/troubleshooting/collector_crash.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Collector crashes
```
lfittl, line 2: :+1: Excellent addition!

lfittl (approved):
Looks great!

A few minor nitpicks, but otherwise I think this is ready to ship!

install/heroku_postgres/01_deploy_the_collector.mdx
uhoh-itsmaciek: Another MDX rendering weirdness--I worked around with a wrapper component like before. But I also noticed we were setting PGA_API_KEY here even if we didn't have it, so I made that part contingent on having the key.

explain/index.mdx
uhoh-itsmaciek: Yeah, that was it. Changed it to use an import. It looks like using relative paths directly for images works with markdown images (``), but not JSX images (``). The latter needs to use imported paths (which also breaks the gatsby-remark image optimization and bells and whistels, but again not a big deal in this case).

PR #31, opened 2021-01-15
Add guided setup instructions

uhoh-itsmaciek:


install.mdx
```diff
@@ -47,7 +47,7 @@ To continue please select in which environment you are running your Postgres dat
 <div className={styles.installChoice}>
   <Link
     className={styles.installChoiceStep}
-    to="install/self_managed/01_create_monitoring_user"
```
lfittl, line 50: We should also update the reference in `index.mdx`

install/self_managed/00_choose_setup_method.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: 'Step 0: Choose Setup Method'
+install_track_title: Installation Guide (Self-Managed)
+backlink_href: /docs/install
+backlink_title: 'Installation Guide'
+---
+
+To monitor a new Postgres server with pganalyze, you must make some configuration changes
+to your database and install and run the pganalyze collector. There are two
+ways of doing this: manually stepping through our instructions, or running the guided
```
lfittl, line 10: This might work better as a list, for better readability (similar to how we present "Log-based EXPLAIN" vs "auto_explain" as two choices)

install/self_managed/01_guided_setup.mdx
```diff
@@ -0,0 +1,44 @@
...
+```
+lsb_release -a
+```
+
+Note that guided setup is supported for Ubuntu 14.04 or newer or Debian 10 or newer. If you are
+running an older version, please follow the manual setup instructions.
+
+If your OS distribution is supported, run the appropriate commands for your version:
+
+<CollectorPkgInstallInstructions kind='deb' />
```
lfittl, line 24: I suspect this should be a follow-up PR, but we should add Ubuntu 20.04 to this.

install/self_managed/01_guided_setup.mdx
```diff
@@ -0,0 +1,44 @@
...
+Guided setup is supported starting in version 0.36.
+
+You'll need to run guided setup as root or using sudo:
+
+```
+sudo pganalyze-collector-setup
+```
+
+You'll be shown additional instructions, and prompted to confirm any changes to your system.
+Once finished, check back here to continue.
```
lfittl, line 44: I think for the public docs we should replace "check back here" with "check the pganalyze app" (or similar)

lfittl (approved):
:+1:

install/self_managed/01_guided_setup.mdx
uhoh-itsmaciek: Yeah, let's do those separately. I'll start an issue.

PR #33, opened 2021-01-21
Refactor log insights setup docs

uhoh-itsmaciek:
This reorganizes the Log Insights setup docs into a series of steps, to make it easier to integrate in-app as online help. I also tried to integrate the "special setups" sections into the appropriate steps, since smaller pages make it easier to handle this inline, and having a separate out-of-band section makes things harder to cross-reference. The troubleshooting sections remain standalone.


uhoh-itsmaciek:
Consolidating into #34

PR #35, opened 2021-02-03
Add API key instructions to guided setup docs.

uhoh-itsmaciek:


install/self_managed/01_guided_setup.mdx
```diff
@@ -34,6 +46,8 @@ pganalyze-collector --version
 
 Guided setup is supported starting in version 0.36.
 
+You will need an API key for the install. <APIKeyInstructions apiKey={props.apiKey} />
+
 You'll need to run guided setup as root or using sudo:
 
 ```
```
lfittl, line 53: I think we have an option we can use here to pass the API key on the command line, correct? (i.e. avoid the prompt and just go directly into the setup with that API key set)

uhoh-itsmaciek: Yeah--I did this because I don't think you can template into the code block, but on second thought I think a better approach is to use the `CodeBlock` wrapper component we are already using and conditionally render the `--api-key` flag there. I'll make that change.

lfittl (approved):
:+1:

PR #37, opened 2021-03-02
Document db_log_syslog_server collector setting

uhoh-itsmaciek:
Not sure if an extra subsection makes sense here, but this is
definitely too much info to fit into the table and all of it is
useful. Thoughts?

### DIFF
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -416,20 +416,11 @@ a self-managed VM on a cloud provider, the configuration settings here may be us
     <tr>
       <td>db_log_syslog_server (<code>LOG_SYSLOG_SERVER</code>)</td>
       <td>[none]</td>
-      <td>local address (host:port) to listen on for syslog messages (see below)</td>
+      <td>local address (host:port) to listen on for syslog messages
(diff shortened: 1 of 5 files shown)


collector/settings.mdx
```diff
@@ -410,12 +410,28 @@ a self-managed VM on a cloud provider, the configuration settings here may be us
...
 
+### Syslog-based Log Insights support
+
+If your database logs are available via syslog, you can use the `db_log_syslog_server` setting to have
+the collector listen for syslog messages from a remote server. This should be a local address like
+"0.0.0.0:32514" and it's recommended to use an unprivileged port number (1024 and up) to avoid running
+the collector as root.
+
+Note that you need to use protocol RFC5424, with an unencrypted TCP connection. Due to syslog not being
+an authenticated protocol it is recommended to only use this integration over private networks.
```
lfittl, line 432: I'm thinking this might be best as its own dedicated docs page - e.g. at `/docs/log-insights/setup/syslog-server`? (and then link it from here, as well as `/docs/log-insights/setup`)

lfittl (approved):

PR #38, opened 2021-03-06
First stab at documenting install script

uhoh-itsmaciek:
This changes our documentation to use the new install script wherever
possible, moves the "manual" package install instructions to a
separate page only in the public docs and links to that page wherever
we reference the script.

Needs testing in-app and we may also want to collapse the
distro-specific collector install pages since many are now identical.

Made the change I had proposed below, so only the manual instructions
include full key information. Other pages just show the fingerprint:

image

---
May also want to avoid linking the key download key on the script pages,
since it may give users the idea that downloading the key is a part of
the install process:

image

Maybe we should just include the fingerprint here and the download only
on the new collector packages page?

### DIFF
--- a/collector/packages.mdx
+++ b/collector/packages.mdx
@@ -1,0 +1,25 @@
+---
+title: 'pganalyze Collector packages'
+backlink_href: /docs/collector
+backlink_title: 'pganalyze Collector'
+---
+
+import CollectorPkgInstallInstructions from '../components/CollectorPkgInstallInstructions'
+
+The collector is available on common Debian-based and Red Hat-based Linux distributions.
+We recommend running our install script to automatically detect your platform and
+install the correct package:
+
+```
+curl https://packages.pganalyze.com/collector-install.sh | bash
+```
+
+Alternately, you can install these packages manually.
+
+<CollectorPkgInstallInstructions />
+
+These repositories are updated by pganalyze whenever a new stable version of the
+collector is released.
+
+You can also build the collector from source: Go 1.16 and `make` are required. To build,
+just run `make`.
--- a/collector/packages.mdx
+++ b/collector/packages.mdx
@@ -6,7 +6,10 @@ backlink_title: 'pganalyze Collector'
 
 import CollectorPkgInstallInstructions from '../components/CollectorPkgInstallInstructions'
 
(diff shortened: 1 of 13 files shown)


collector/packages.mdx
```diff
@@ -0,0 +1,25 @@
+---
+title: 'pganalyze Collector packages'
+backlink_href: /docs/collector
+backlink_title: 'pganalyze Collector'
+---
+
+import CollectorPkgInstallInstructions from '../components/CollectorPkgInstallInstructions'
+
+The collector is available on common Debian-based and Red Hat-based Linux distributions.
```
lfittl, line 9: Maybe "Pre-built collector packages are available on common Debian-based and Red Hat-based Linux distributions for x86-based architectures. Support for BSD-based platforms, or other architectures like ARM is available by [building from source](https://github.com/pganalyze/collector)."

collector/packages.mdx
```diff
@@ -0,0 +1,25 @@
...
+```
+
+Alternately, you can install these packages manually.
+
+<CollectorPkgInstallInstructions />
+
+These repositories are updated by pganalyze whenever a new stable version of the
+collector is released.
+
+You can also build the collector from source: Go 1.16 and `make` are required. To build,
```
lfittl, line 24: We could also link "from source" to https://github.com/pganalyze/collector here

install/self_managed/01_guided_setup.mdx
```diff
@@ -5,44 +5,27 @@ backlink_href: /docs/install
...
+import CollectorInstallInstructions from '../../components/CollectorInstallInstructions'
 
 export const GuidedSetupInvocationExample = ({ apiKey }) => {
+  const installEnv = { PGA_GUIDED_SETUP: 'true' }
+  if (!!apiKey) {
+    installEnv['PGA_API_KEY'] = apiKey
+  }
   return (
-    <CodeBlock>sudo pganalyze-collector-setup{apiKey && ` --api-key=${apiKey}`}</CodeBlock>
+    <CollectorInstallInstructions env={installEnv} />
```
lfittl, line 16: Nice!

lfittl (approved):
:+1:

PR #39, opened 2021-03-15
Recommend guided setup for broader usage

uhoh-itsmaciek:
Remove the suggestion to avoid using guided setup on production
systems to encourage broader adoption.


lfittl (approved):
:+1:

PR #41, opened 2021-04-09
Move Check docs into docs repo

uhoh-itsmaciek:
This does not yet include public check documentation, but allows the
core of the documentation to be shared.

### DIFF
--- a/components/CheckDocumentation/connections/ActiveQuery.tsx
+++ b/components/CheckDocumentation/connections/ActiveQuery.tsx
@@ -1,0 +1,135 @@
+import React from "react";
+
+import {
+  CheckDocs,
+  CheckGuidanceProps,
+  CheckTriggerProps,
+  IssueReferenceBackend,
+} from "../../../util/checks";
+
+import PGDocsLink from "../../PGDocsLink";
+import SQL from "../../SQL";
+import { useSmartAnchor } from "../../SmartAnchor";
+
+const ActiveQueryTrigger: React.FunctionComponent<CheckTriggerProps> = ({
+  config,
+}) => {
+  return (
+    <>
+      <p>
+        Detects queries currently running for longer than the specified
+        threshold of{" "}
+        <code>{config.settings["warning_max_query_age_secs"]}</code> seconds and
+        creates an issue with severity "warning". Escalates to "critical" if the
+        query is still running after{" "}
+        <code>{config.settings["critical_max_query_age_secs"]}</code> seconds.
+        Resolves automatically once the query stops running.
+      </p>
+      <p>
+        Ignores queries from backup programs (<code>pg_dump</code>,{" "}
+        <code>Heroku Postgres Backups</code>), as well as any queries that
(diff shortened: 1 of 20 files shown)


components/CheckDocumentation/connections/ActiveQuery.tsx
```diff
@@ -0,0 +1,135 @@
...
+      </p>
+    </>
+  );
+};
+
+const ActiveQueryGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
+  urls: { referenceUrl, serverVacuumsUrl, SettingLink },
+  issue,
+}) => {
+  const Link = useSmartAnchor();
```
uhoh-itsmaciek, line 42: These are going to be hard to diff because they're new to this repo, but this (across all docs components) is the only change other than imports.

components/PGDocsLink/index.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { useIcon } from "../WithIcons";
+
+type Props = {
+  path: string;
+  version?: string;
+};
+
+const PGDocsLink: React.FunctionComponent<Props> = ({
```
uhoh-itsmaciek, line 9: I think it makes sense to just move this component here wholesale.

components/SQL/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  "position",
+  "is",
+  "id",
+  "cleanup",
+  "query",
+  "extract",
+  "coalesce",
+];
+
+class SQL extends React.Component<Props> {
```
uhoh-itsmaciek, line 36: Copying this component in here because it's clearly needed in docs, but it's also needed in-app for non-documentation issues. The right solution is, as we've discussed, a separate shared component package, but that complicates things a bit. I think a copy is okay for now, but open to other ideas.

components/SmartAnchor.tsx
```diff
@@ -66,6 +73,27 @@ const SmartAnchor: React.FunctionComponent<Props> = ({linkComponent, linkRelativ
...
+  linkRelative: boolean;
+}> = ({linkComponent, linkRelative, children}) => {
+  return (
+    <SmartAnchorContext.Provider value={makeSmartAnchor(linkComponent, linkRelative)}>
+      {children}
+    </SmartAnchorContext.Provider>
+  )
+}
+
+export function useSmartAnchor(): React.ComponentType<BaseProps> {
```
uhoh-itsmaciek, line 89: With MDX, we can use shortcodes to inject new component types globally, but since check documentation is plain React, not MDX, we need to use this injection mechanism (we're doing something similar for icons).

components/WithIcons.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useContext } from 'react'
 import Null from './Null';
 
 export type IconProps = {
-  className: string;
+  className?: string;
```
uhoh-itsmaciek, line 5: The `PGDocsLink` does not need a custom `className` so it seems reasonable to make this optional.

lfittl (approved):
👍

PR #42, opened 2021-04-12
Tweaks to support check docs in public documentation

uhoh-itsmaciek:
Some further changes necessary to make the check docs available in
public documentation.


checks/index.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+To help you discover potential problems, pganalyze automatically performs a number
+of checks on your database.
+
+Checks run regularly, at different frequencies depending on the check type. When
+a problem is detected, an issue is created for the server or database where the
+problem is found.
+
+Issues may have severity "info", "warning", or "critical". Critical issues may
+cause a serious performance problem or lead to a database outage within a matter
```
lfittl, line 43: Maybe better if worded like this:

"Critical issues may indicate a serious performance problem or potential database outage within a matter of hours."

(i.e. to separate the issue in pganalyze from the actual problem, since e.g. if you're getting a lot of active query alerts that itself may not be the root cause)

checks/index.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+
+## Alerting
+
+If an issue is of severity "critical" (either because it starts as critical or
+escalates to that level), it will alert organization users according to their
+configured Alert Policy. You can edit Alert Policies in the Configure section
+of the Alerts & Check-Up page. You can edit organization members' assigned
+policies in each member's organization profile page, accessible from the Members
+section of the organization Settings page.
```
lfittl, line 56: :+1:

checks/index.mdx
```diff
@@ -0,0 +1,63 @@
...
+            </ul>
+          </li>
+        )
+      })}
+    </ul>
+  )
+}
+
+To help you discover potential problems, pganalyze automatically performs a number
+of checks on your database.
```
lfittl, line 36: I wonder if we should include a screenshot here, similar to how we do on the Log Insights and automatic EXPLAIN pages. For example this one:

Screen Shot 2021-03-31 at 12 51 16 PM

util/checks/index.ts
```diff
@@ -93,8 +179,16 @@ export type IssueGuidanceUrls = {
   databaseTableUrl: string;
   serverLogInsightsUrl: string;
   serverSchemaUrl: string;
+  sectionUrl: (mainUrl: string | undefined, section: string) => string | undefined;
 };
 
+export function sectionUrl (mainUrl: string | undefined, section: string): string | undefined {
```
lfittl, line 185: I wonder if this should maybe be called "featureUrl" instead? (since it points to a particular feature in-app, and to a page describing that feature in the public docs)

lfittl (approved):
:+1:

checks/index.mdx
uhoh-itsmaciek: Great call.

checks/index.mdx
uhoh-itsmaciek: Yeah, good idea. I'll just use this screenshot.

util/checks/index.ts
uhoh-itsmaciek: Sure. I considered that but thought it was toss-up. I'll change to 'featureUrl'.

PR #43, opened 2021-04-16
Add Sort Key to EXPLAIN Sort nodes important fields

uhoh-itsmaciek:
This will have it show in the Explain Node Details panel.


lfittl (approved):
👍

PR #47, opened 2021-05-06
Collector settings: Add separate section for schema filter settings

lfittl:
This may even justify its own docs page in the future, but for now
giving it its own section makes it easier to link to, and allows adding
some more explanation and a test query to verify whether the setting is
working as intended.


uhoh-itsmaciek (approved):
:+1:, this works. In the future, it might be nice to have the collector test tell you monitored object counts, but let's start with this.

PR #48, opened 2021-05-10
Add documentation for aws_web_identity_token_file and aws_role_arn

uhoh-itsmaciek:
These settings are handy for configuring AWS access when running in
EKS.

My descriptions are pretty vague so I'm open to improvements, but we
should document this. Should we link the actual
AssumeRoleWithWebIdentity docs [1]? Other thoughts?

[1] https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html


lfittl (approved):
:+1:

Nice, thanks for making the effort to add the docs!

I think the wording as is works well :)

PR #50, opened 2021-05-26
Heroku setup steps: Fix typo, ensure add-on attach uses underscores

lfittl:


install/heroku_postgres/util.ts
```diff
@@ -2,5 +2,5 @@ export const collectorAppName = (organizationSlug: string) =>
   `${organizationSlug || "testapp"}-pganalyze-collector`;
 export const attachmentName = (organizationSlug: string) =>
   organizationSlug
-    ? `${organizationSlug.toUpperCase()}_PRIMARY`
+    ? `${organizationSlug.toUpperCase().replace("-", "_")}_PRIMARY`
```
uhoh-itsmaciek, line 5: Maybe `.replace(/[^a-zA-Z0-9]+/g, '_')`? I think those are the only characters allowed in config vars (plus the underscore, but omitting the underscore in the regexp means it would change "foo-_-bar" to "foo_bar" which I think makes sense).

Even if we want to stick with just filtering out dashes, that should be `.replace(/-/g,"_")`--the string replace will only replace the first occurrence.

uhoh-itsmaciek (changes requested):

install/heroku_postgres/util.ts
uhoh-itsmaciek: @lfittl let me know if you want me to take over and wrap this up--just ran into this myself setting up the demo org for PD :grimacing:

uhoh-itsmaciek: There's some other feedback on Heroku docs so I'll just roll these changes into that.

lfittl: @uhoh-itsmaciek Yeah, that sounds good - feel free to take this over. Thanks :)

uhoh-itsmaciek:
Superseded by #55

PR #53, opened 2021-06-09
Add PagerDuty integration docs

uhoh-itsmaciek:
Following the template provided at https://github.com/PagerDuty/app-documentation-templates/blob/master/integration-guide-template.md

### DIFF
--- a/checks/pagerduty.mdx
+++ b/checks/pagerduty.mdx
@@ -50,7 +50,7 @@ Before configuring alerts, you will need to connect your pganalyze organization 
 3. On the organization settings page, click the **Integrations** tab
 4. In the panel labeled **PagerDuty**, click the **Connect to PagerDuty** button
 5. Continue to authenticate with PagerDuty and select which services you would like to connect to pganalyze
-   * services you select will later be available to configure for alerting; no alerting is enabled by default
+   * Services you select will later be available to configure for alerting; no alerting is enabled by default
 6. Once connected, select a service from the dropdown and click **Test integration** to trigger a test alert
    to the selected service and verify that the integration is configured correctly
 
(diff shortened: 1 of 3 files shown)


checks/pagerduty.mdx
```diff
@@ -0,0 +1,91 @@
...
+### Initial setup
+
+Before configuring alerts, you will need to connect your pganalyze organization to PagerDuty:
+
+1. In the top menu, click the **x** in the **Server** dropdown if a server is selected to clear the active server
+2. In the left menu, click the **Settings** item
+3. On the organization settings page, click the **Integrations** tab
+4. In the panel labeled **PagerDuty**, click the **Connect to PagerDuty** button
+5. Continue to authenticate with PagerDuty and select which services you would like to connect to pganalyze
+   * services you select will later be available to configure for alerting; no alerting is enabled by default
```
lfittl, line 53: I think we can uppercase `services` here, since it's at the start of a sentence.

checks/pagerduty.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+### Configuring alerts
+
+1. In the top menu, use the **Server** dropdown to select the server to configure
+2. In the left menu, click the **Alerts & Check-Up** item
+3. On the server alerts page, click the **Configure** tab
+4. Click the **Configure alerts for this server** link
+5. Select a service from the **PagerDuty service** dropdown
+6. Check the checkboxes in the **PagerDuty Alert** column for the checks you would like to enable
+   * only checks which escalate to critical severity will be selectable
```
lfittl, line 66: Same here, I think `only` could be `Only`

lfittl (approved):
Nice, that reads well!

I think we could consider including a screenshot or two, but up to you - I think this is what's necessary for now to get the integration approved.

uhoh-itsmaciek:
The PagerDuty docs counsel against screenshots on the grounds they're harder to update as things change. Not sure I buy that entirely, but it's not unreasonable. I'd rather follow their guidance on this, though.

checks/pagerduty.mdx
uhoh-itsmaciek: Sure--I thought as a sub-item this could work lowercase, but you're right, when rendered it reads better capitalized.

PR #61, opened 2021-09-21
Add initial Index Advisor docs

uhoh-itsmaciek:


index-advisor/standalone/index.mdx
```diff
@@ -0,0 +1,100 @@
...
+   - uuid-ossp
+
+## Common Errors
+
+Many of the errors you are likely to run into in Index Advisor output
+are due to the limitations outlined above. Please review the limitations
+and, if applicable, try the workarounds we suggest.
+
+ - Unexpected execution of function 
+   - Custom functions are not supported.
```
lfittl, line 94: Technically this can happen for built-in functions as well, since we have an allowlist for which built-in functions we allow you to call. Maybe reword: "The index advisor planner executed a function unexpectedly. If this is a built-in function provided by Postgres, [let us know](feedback link). If this error occurs for a custom function you created this is an unsupported scenario."

lfittl (approved):
:+1:

PR #63, opened 2021-10-18
Make docs compatible with MDX 2

uhoh-itsmaciek:
As part of the move to Vite.js, we need to upgrade in-app doc
rendering to @mdx-js/mdx v2, which seems to have some minor rendering
differences:

 - Some top-level React elements in MDX now always appear to render as
   display:block, so work around that with a wrapper element.
 - The '<>' shorthand for React.Fragment is no longer recognized. This
   will probably be fixed at some point, but for now, just use the long
   form.
 - The Postgres logo in the "select environment" overview is now very small
   (possibly since this is the only cell that stacks three lines); force
   dimensions similar to what we had before.


lfittl (approved):
:+1:

PR #69, opened 2021-11-18
Add index recommendation URL to issue guidance URLs

uhoh-itsmaciek:
This lets us link to the index advisor when a plan supports it or to
the index check otherwise.

### DIFF
--- a/components/CheckDocumentation/queries/Slowness.tsx
+++ b/components/CheckDocumentation/queries/Slowness.tsx
@@ -28,6 +28,7 @@ const SlownessGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
   urls: {
     firstReferenceUrl,
     queriesUrl,
+    indexRecommendationUrl,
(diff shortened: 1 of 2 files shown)


lfittl (approved):
Also just to record my own understanding: In the public docs (https://pganalyze.com/docs/checks/queries/slowness) this is already just text (not a link), so we don't need to do anything there.

uhoh-itsmaciek:
@lfittl exactly:

```typescript
export function featureUrl (mainUrl: string | undefined, section: string): string | undefined {
  if (mainUrl == undefined) {
    return undefined;
  }
  return `${mainUrl}/${section}`;
}
```

and the `Link` component we use comes from `const Link = useSmartAnchor();`, which is configured to use SmartAnchor in the public docs, which accounts for that here: https://github.com/pganalyze/pganalyze-docs/blob/main/components/SmartAnchor.tsx#L35-L38

PR #74, opened 2021-12-08
Upgrade to highlight.js 11 to fix errors

uhoh-itsmaciek:
We upgraded highlight.js to 10 during our Gatsby upgrade and something
broke. Since 11 appears to work with only minor changes, update the
imports to support that.


lfittl (approved):
:+1:

PR #79, opened 2021-12-24
Guided setup: Remove link to manual pkgs, instead point to manual setup

lfittl:
The main choice that one makes here is "guided" vs "manual" setup,
and the page offered no clear way to go back to the manual setup
when exploring the guided setup option.

Additionally, there was a confusingly worded text adjacent to the install
script. Since we don't describe a way currently to run the guided setup
with a manual package install, remove that link altogether for the guided
setup page.

In passing, add an emphasis on guided setup only support Ubuntu/Debian.


uhoh-itsmaciek (approved):
Sorry I missed this (there was no review request, so I was not notified). Looks good.

lfittl:
> Sorry I missed this (there was no review request, so I was not notified). Looks good.

Thanks, I had forgotten about this as well 🙃

PR #80, opened 2022-01-05
Update EXPLAIN set up docs to use auto_explain for GCP

uhoh-itsmaciek:


explain/setup/google_cloud_sql/01_auto_explain_check.mdx
```diff
@@ -0,0 +1,21 @@
...
+---
+
+import Step01InApp from './_01_in_app.mdx'
+import Step01Public from './_01_public.mdx'
+
+The recommended Automated EXPLAIN configuration for Google Cloud SQL is to use
+the Postgres `auto_explain` contrib module. This ensures accurate and detailed plan
+information, including runtime information (as gathered by EXPLAIN ANALYZE). With our
+recommended configuration (presented in a later step), performance overhead is minimal.
+
```
lfittl, line 15: It looks like the step components included have RDS-specific wording, probably good to fix:

"First, to check if auto_explain is already enabled on your system, run the following query as an **RDS superuser**:"

explain/setup/google_cloud_sql/04_test_and_verify.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+```
+sudo pganalyze-collector --test-explain
+```
+
+This will produce no output but will run an artificially slow query using the
+Postgres `pg_sleep` function.
+
+<PublicOnly>
+  <VerifyExplainPublic />
```
lfittl, line 25: Not a big deal, but I noticed the screenshot we have on this step has our old navigation on the left side - might be nice to fix so it matches what you'd actually see.

explain/setup/index.mdx
```diff
@@ -34,7 +34,7 @@ Platform         | Log-based EXPLAIN |    auto_explain   |
 -----------------|-------------------|-------------------|
 Amazon RDS       | Yes               | Yes (Recommended) |
 Azure Database   | Yes               | No                |
```
lfittl, line 36: Not urgent, but I just realized we haven't updated the Azure docs to reflect that Azure also supports auto_explain now. We can discuss separately how we want to prioritize that.

lfittl (approved):
This is excellent, and thanks for testing this and adding images - that's very helpful!

Apart from some minor comments I noticed one stale reference to log-based EXPLAIN on `/docs/log-insights/setup/google-cloud-sql/04_configure_collector` - that should probably talk about auto_explain now as well (and link to the auto_explain docs).

explain/setup/google_cloud_sql/01_auto_explain_check.mdx
uhoh-itsmaciek: This is actually an instance of the stupid Gatsby MDX behavior I was complaining about. The components are right, but because they share the same name (in totally different files!) there's a collision and the wrong one is pulled in. Good catch.

lfittl: Oh. Sigh. *yells at cloud*

explain/setup/google_cloud_sql/04_test_and_verify.mdx
uhoh-itsmaciek: I can update these :+1:.

explain/setup/index.mdx
uhoh-itsmaciek: Ah, I didn't even know Azure had auto_explain, too. But yeah, let's do that separately to avoid holding this up.

PR #81, opened 2022-01-14
Update RDS docs settings links to point to official docs

uhoh-itsmaciek:
Right now, these point to demo.pganalyze.com, but due to the
navigation changes from 64f2fda29f88f52e022fc27a9c7fd550063b0a6f and
follow up commits, these routes are not valid.

Since we've been de-emphasizing demo.pganalyze.com, point these to
the official Postgres docs rather than fixing the links.


seanlinsley (approved):

lfittl (approved):
👍

PR #82, opened 2022-01-18
Add note about DEFAULT_ORG_ROLE to LDAP instructions

uhoh-itsmaciek:
If one initially sets up an organization using a password-based login,
but then enables LDAP, the LDAP users will not be members of the
organization (and will not be able to create new organizations due to
Enterprise restrictions).

Include a note explaining this in the LDAP instructions.

### DIFF
--- a/enterprise/setup/index.mdx
+++ b/enterprise/setup/index.mdx
@@ -282,9 +282,9 @@ AWS_REGION=us-east-1
 
 ### Appendix D - LDAP configuration
 
-If configuring LDAP after initially setting up an organization with standard password-based login,
-be sure to set DEFAULT_ORG_ROLE (see above) to ensure that new LDAP users are added to the existing
-organization.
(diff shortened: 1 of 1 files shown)


enterprise/setup/index.mdx
```diff
@@ -282,6 +282,10 @@ AWS_REGION=us-east-1
 
 ### Appendix D - LDAP configuration
 
+If configuring LDAP after initially setting up an organization with standard password-based login,
+be sure to set DEFAULT_ORG_ROLE (see above) to ensure that new LDAP users are added to the existing
+organization.
```
lfittl, line 287: I think we can actually state this more generally, as you'd run into issues here regardless of how the organization is set up:

```
When using LDAP, it is recommended to also set DEFAULT_ORG_ROLE to a non-standard value (see above) to ensure that new LDAP users are added to the existing organization.
```

lfittl (approved):
👍  (with a rewording)

PR #84, opened 2022-02-08
Add Log Insights rate limit documentation

uhoh-itsmaciek:


log-insights/limits.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+As the collector sends Log Insights data to pganalyze for processing, the service tracks
+the total number of bytes received for each server. Each server starts with a 1 GB quota
+that is continuously refilled at the rate of 1 GB per day.
+
+Effectively, each server has a 1 GB allotment of log data over a rolling 1 day period. This
+allows for sending bursts of log data as long as the average rate is under 1 GB per day.
+
+When incoming log data exceeds the limit for a particular server, the affected log lines
+are not recorded by pganalyze and will not show up in Log Insights. Query samples present
```
lfittl, line 23: Maybe "EXPLAIN plans and query samples" (in the UI we have these as separate tabs)

log-insights/limits.mdx
```diff
@@ -0,0 +1,43 @@
...
+in affected lines are still recorded. The amount of log lines dropped, in bytes, is
+recorded whenever the rate limit is exceeded. This is visible as a warning when you visit
+the Log Insights page for an affected server.
+
+All Postgres logs processed by the collector count toward this limit.
```
lfittl, line 28: Not sure if the extra "Postgres" here isn't confusing (vs helpful), since it might imply that there is another kind of log the collector tracks that don't count towards the limit.

log-insights/limits.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## Settings to adjust to stay under the rate limit
+
+There are a number of [Postgres log settings](https://www.postgresql.org/docs/current/runtime-config-logging.html)
+that can be adjusted to help you stay under this threshold:
+
+ - Disable `log_duration` if set
```
lfittl, line 35: This makes me think, on this page, should we also talk about the settings where the collector just doesn't collect logs at all and puts a warning instead? (e.g. "log_statement = all", "log_min_duration_statement" under 10ms, etc)

lfittl (approved):

log-insights/limits.mdx
```diff
@@ -0,0 +1,43 @@
+---
+title: 'pganalyze Log Insights Limits'
+backlink_href: /docs/log-insights
+backlink_title: 'Log Insights'
+---
+
+Log Insights has a rate limit of 1 GB of log data per day for each server. Limits for
```
lfittl, line 7: Should we hint at increased limits being available by contacting us? (we have a rather basic way to do this today, but if there is a larger interest in it, we could make this a self-serve feature)

e.g. "Additional log limit quota can be added to your plan by contacting pganalyze Support.".

log-insights/limits.mdx
uhoh-itsmaciek: Yeah, I was on the fence. I'll strike it.

log-insights/limits.mdx
uhoh-itsmaciek: Sure :+1:

log-insights/limits.mdx
uhoh-itsmaciek: Sure, I'll add a section for this.

PR #87, opened 2022-03-07
Clarify references to API keys in-app

uhoh-itsmaciek:
Currently, we talk about the API Keys page, but it may not be obvious
how to get there.

Describe the full navigation necessary to get to the API keys page
(Settings page, API keys tab) when referencing it.


uhoh-itsmaciek:
This could also probably be unified to avoid copying this text all over the place, but it's not quite the same everywhere (sometimes for a good reason), so I think it's a separate problem.

lfittl (approved):
👍

PR #88, opened 2022-03-31
Add Crunchy Bridge documentation

lfittl:


components/MonitoringUserSetupInstructions.tsx
```diff
@@ -29,30 +28,125 @@ const MonitoringUserSetupInstructions: React.FunctionComponent<Props> = ({
...
+  );
+}
+
 const MonitoringUser10: React.FunctionComponent<{password: string}> = ({password}) => {
   const CodeBlock = useCodeBlock();
   return (
     <CodeBlock>
       {`CREATE USER pganalyze WITH PASSWORD '${password}' CONNECTION LIMIT 5;
 GRANT pg_monitor TO pganalyze;
-
```
uhoh-itsmaciek, line 55: If you think we should drop this line break, maybe we should do it consistently across the older versions? Otherwise this will jump for no reason when users are switching tabs.

Also, we should probably drop the 9.4 instructions (but we can do that separately).

uhoh-itsmaciek (approved):

components/MonitoringUserSetupInstructions.tsx
lfittl: That was an accident, will remove this change & then merge (and agreed on 9.4).

PR #91, opened 2022-04-06
Update SSO docs

uhoh-itsmaciek:
Try to make the SSO instructions more explicit, and focused on the
(presumably) common use case of migrating all users to SSO once the
feature is enabled.


lfittl (approved):
:+1:

PR #94, opened 2022-05-23
Add troubleshooting instructions for setting up column stats helper

uhoh-itsmaciek:
image

### DIFF
--- a/install/troubleshooting/column_stats_helper.mdx
+++ b/install/troubleshooting/column_stats_helper.mdx
@@ -1,0 +1,28 @@
+---
+title: 'Resolving the "Limited access to table column statistics" warning'
+backlink_href: /docs/install/troubleshooting
+backlink_title: 'Installation Troubleshooting'
+---
+
+import { MonitoringUserColumnStats } from "../../components/MonitoringUserSetupInstructions";
+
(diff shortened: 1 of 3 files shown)


install/troubleshooting/column_stats_helper.mdx
```diff
@@ -0,0 +1,28 @@
...
+```
+
+This warning indicates that column statistics for the specific database are not being reported by the collector.
+Column statistics are used by pganalyze to improve Index Advisor recommendations. Index Advisor will still work
+without the helper functions, but index recommendations may be less accurate.
+
+To resolve this, find the name for the database role that's configured as the monitoring user
+in your collector installation (we'll assume "pganalyze" here).
+
+<MonitoringUserColumnStats username="pganalyze" />
```
lfittl, line 22: Nice!

lfittl (approved):
:+1:

install/troubleshooting/column_stats_helper.mdx
uhoh-itsmaciek: Yeah, it was nice to reuse this. I had to get a little creative with the copy since the first paragraph in this component starts with "Then, ...", but I think it works.

PR #98, opened 2022-06-24
Add Aiven setup instructions

uhoh-itsmaciek:
Also add Log Insights setup instructions, and link to existing
log-based EXPLAIN instructions.


components/MonitoringUserSetupInstructions.tsx
```diff
@@ -103,8 +116,8 @@ $$ LANGUAGE plpgsql VOLATILE SECURITY DEFINER;`}
         data: it ensures that the collector cannot piggyback other queries that could exfiltrate data.
       </p>
       <p>
-        Because EXPLAIN needs to run in the same database where the query ran, you will need to create this function
-        <strong>in each database you want to monitor</strong>. You should create the function as a Postgres superuser to ensure EXPLAIN
+        Because EXPLAIN needs to run in the same database where the query ran, you will need to create this function <strong>in each
```
uhoh-itsmaciek, line 119: Because of JSX whitespace rules, this rendered as "...function**in each database**...", so just fixing that.

explain/setup/log_explain/01_create_helper_functions.mdx
```diff
@@ -22,5 +22,5 @@ The safest way to permit this is to create the following helper function **in ea
 <MonitoringUserExplain username="pganalyze" />
 
 <Link className='btn btn-success' to='02_enable_log_explain'>
-  Proceed to Step 2: Configure collector
+  Proceed to Step 2: Enable log EXPLAIN
```
uhoh-itsmaciek, line 25: This had the wrong link text.

install.mdx
```diff
@@ -83,6 +74,25 @@ To continue please select in which environment you are running your Postgres dat
     EnterpriseDB
   </Link>*/}
```
uhoh-itsmaciek, line 75: Theoretically we should be able to remove these commented-out tiles, but removing them breaks all images on the page. :confounded:

log-insights/setup/aiven/03_configure_collector.mdx
```diff
@@ -0,0 +1,55 @@
...
+```
+
+(If using Docker, use the `LOG_SYSLOG_SERVER` environment variable instead.)
+
+It's recommended to use an unprivileged port number (1024 and up) to avoid
+running the collector as root.
+
+Restart the collector.
+
+Make sure that the port you've selected is not blocked by a firewall or security group rules.
```
lfittl, line 31: Should we clarify here that this is for a TCP connection? (since adding the firewall rule / security group rule requires you to pick whether its UDP or TCP)

log-insights/setup/aiven/03_configure_collector.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+export const ImgAwsSecurityGroup = () => <img alt="AWS Security Group allowing access on syslog port" src={imgAwsSecurityGroup} />
+
+We can now configure the syslog receiver in the collector. Edit the collector
+config and add the
+[`db_log_syslog_server`](https://pganalyze.com/docs/collector/settings#self-managed-servers)
+setting. This should be set to a local address on any free port:
+
+```
+db_log_syslog_server: 0.0.0.0:32514
```
lfittl, line 21: Since the screenshots below use port `50014`, should we use that here as well for consistency?

log-insights/setup/aiven/03_configure_collector.mdx
```diff
@@ -0,0 +1,55 @@
...
+<ImgGcpFirewallRules />
+
+In AWS, you'll need to add a security group rule:
+
+<ImgAwsSecurityGroup />
+
+You can verify that the port is accessible with telnet:
+
+```
+$ telnet 203.0.113.2 50014
```
lfittl, line 43: Since we talk about the VPC peering and keeping things private earlier, its a bit odd that we are now referencing a public IP here (and a specific IP as well?)

maybe we should clarify which IP we mean here? (presumably the public IP of where the collector runs)

log-insights/setup/aiven/03_configure_collector.mdx
```diff
@@ -0,0 +1,55 @@
...
+Make sure that the port you've selected is not blocked by a firewall or security group rules.
+E.g., in GCP, you'll need to add a firewall rule:
+
+<ImgGcpFirewallRules />
+
+In AWS, you'll need to add a security group rule:
+
+<ImgAwsSecurityGroup />
+
+You can verify that the port is accessible with telnet:
```
lfittl, line 40: Not a big deal, but on macOS I have `nc` but not `telnet` by default - maybe we should be using `nc` here?

```
lfittl@starfish pganalyze % nc -v 107.21.62.12 32514
Connection to 107.21.62.12 port 32514 [tcp/*] succeeded!
^C
lfittl@starfish pganalyze % nc -v 107.21.62.12 32515
^C
```

(if we keep telnet we may want to link https://superuser.com/questions/486496/how-do-i-exit-telnet -- as I had forgotten too how I would produce `^]`..)

log-insights/setup/aiven/04_configure_service_integration.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+Select the endpoint you just created from the dropdown and click "Enable". Then
+dismiss the integrations screen.
+
+After a few minutes, the Service Integrations item should refresh automatically and
+show your integration as active:
+
+<ImgSyslogIntegrationActive />
+
+<PublicOnly>
```
lfittl, line 53: Comment on the in-app experience for the spinner here - the notion of "once you've run the test command" is a bit odd, maybe we should special case that for platforms where we're streaming instead of downloading:

Screen Shot 2022-07-11 at 7 13 34 PM

lfittl (approved):
:+1: Works as expected!

components/MonitoringUserSetupInstructions.tsx
```diff
@@ -261,5 +276,65 @@ $$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}
...
+  const CodeBlock = useCodeBlock();
+  return (
+    <>
+      <p>
+        Because your Postgres provider does not give you access to the <code>pg_monitor</code>
+        role that can read all queries from pg_stat_statements, you'll need to create a set of
+        SECURITY DEFINER helper functions to collect full query stats.
+      </p>
+      <p>
+        For each user whose queries you would like to monitor, you'll need to log in to your
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Maybe. I tried to keep this parallel with the self-managed instructions for easier maintenance. I'll review what we can do here.

install/aiven/04_configure_the_collector_package.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+Fill in the values from the info in your Aiven console:
+
+<ImgAivenDbSettings />
+
+1. <APIKeyInstructions apiKey={props.apiKey} />
+2. The `db_host` is the hostname of your PostgreSQL instance (`example.aivencloud.com` above)
+3. The `db_port` is the port your PostgreSQL server is running on (`12448` above)
+4. The `db_name` is the database you want to monitor (`defaultdb` above)
+5. The `db_username` and `db_password` should be the credentials of the monitoring user we created in Step 1
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Yeah, I've thought about that but haven't seen any support tickets. I don't think just a button is much of a UX improvement. We could maybe cache it short-term while you are in the same install track? In any case, since it's a preexisting issue and doesn't have a simple fix, I think we should address it separately.

install/aiven/04_configure_the_collector_package.mdx
```diff
@@ -0,0 +1,67 @@
...
+export const ConfigSample = ({ apiKey }) => {
+  return (
+    <CodeBlock>
+      {`[pganalyze]
+api_key = ${apiKey ?? 'your_pga_organization_api_key'}\n
+[server1]
+db_host = your_database_host
+db_port = your_database_port
+db_name = your_database_name
+db_username = your_monitoring_user
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Yeah, we could standardize on 'pganalyze' in examples, but I have seen support interactions with monitoring users with other names, so that's not necessarily a clear win. In either case, I think we should stay consistent with other instructions for now and address this separately.

log-insights/setup/aiven/03_configure_collector.mdx
uhoh-itsmaciek: Ah, `nc` is a  great idea.

log-insights/setup/aiven/04_configure_service_integration.mdx
uhoh-itsmaciek: Good point, but also a preexisting issue for other platforms; let's handle it separately.

install/aiven/04_configure_the_collector_package.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+export const ConfigSample = ({ apiKey }) => {
+  return (
+    <CodeBlock>
+      {`[pganalyze]
+api_key = ${apiKey ?? 'your_pga_organization_api_key'}\n
+[server1]
+db_host = your_database_host
+db_port = your_database_port
+db_name = your_database_name
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Maybe, although if you've changed it in the past, that might be confusing. I think I'd rather stick with consistent instructions and address this if it ends up being a problem in support.

log-insights/setup/aiven/03_configure_collector.mdx
uhoh-itsmaciek: That's a great point. I think this was from a note I made when I was debugging and was just trying to verify that a public internet connection worked. But if we want to encourage a secure setup, we don't want this to be accessible over the public internet at all (no one can snoop on your log traffic if that connection is using the VPC peering, but you could still get a DoS or spam with a setup like that). Ideally, we'd want you to verify that you can connect from the Aiven VPC , but you can't really run `nc` there. Maybe we should just drop this altogether?

The IP I chose is from https://datatracker.ietf.org/doc/html/rfc5737, but yeah maybe that's surprising if you don't know about it.

lfittl: Yeah, maybe dropping it is fine - this isn't really specific to our setup (though helpful to test security group settings).

components/MonitoringUserSetupInstructions.tsx
```diff
@@ -261,5 +276,65 @@ $$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}
...
+      </p>
+      <CodeBlock>
+        {`CREATE OR REPLACE FUNCTION pganalyze.get_stat_statements_analytics(showtext boolean = true)
+RETURNS SETOF pg_stat_statements AS
+$$
+  SELECT * FROM public.pg_stat_statements(showtext) WHERE userid = 'analytics'::regrole;
+$$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}
+      </CodeBlock>
+      <p>
+        Then, log in as {adminUserStr} and create a helper function to combine these. For example,
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Good point--clarified this.

components/MonitoringUserSetupInstructions.tsx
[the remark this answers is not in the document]
uhoh-itsmaciek: Move that to its own step, and it flows better in addition to avoiding errors. Thanks for catching this. I originally moved this up after going through setup and thought it was safe because the function body is not evaluated until later, but the return type depends on the extension, too. Having this as its own step works better.

install/aiven/02_enable_pg_stat_statements.mdx
```diff
@@ -0,0 +1,46 @@
...
+```
+
+If you have configured your database correctly, this will return a result like this:
+
+```
+ calls | query
+-------+-------
+     8 | SELECT * FROM t WHERE field = ?
+(1 row)
+```
```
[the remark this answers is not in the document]
uhoh-itsmaciek: Updated to do that.

PR #99, opened 2022-06-24
Fix escaping in helper function documentation

uhoh-itsmaciek:
In two places, we're using regular expressions with backslashes, but
because these are in TypeScript backtick-quoted strings, the
backslashes are interpreted as escapes for the following
characters. Escaped characters with no special meaning render as
themselves, but this means the backslashes are stripped.

Add additional backslashes to escape the backslashes we want to keep.

### DIFF
--- a/components/MonitoringUserSetupInstructions.tsx
+++ b/components/MonitoringUserSetupInstructions.tsx
@@ -74,7 +74,7 @@ DECLARE
   prepared_params text;
   result text;
 BEGIN
-  SELECT regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
+  SELECT regexp_replace(query, ';+\\s*\\Z', '') INTO prepared_query;
   IF prepared_query LIKE '%;%' THEN
     RAISE EXCEPTION 'cannot run EXPLAIN when query contains semicolon';
   END IF;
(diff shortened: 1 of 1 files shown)


components/MonitoringUserSetupInstructions.tsx
```diff
@@ -74,7 +74,7 @@ DECLARE
   prepared_params text;
   result text;
 BEGIN
-  SELECT regexp_replace(query, ';+\s*\Z', '') INTO prepared_query;
+  SELECT regexp_replace(query, ';+\\s*\\Z', '') INTO prepared_query;
```
lfittl, line 77: Huh, interesting. I wonder if this has caused problems in the past where people have a bad function definition.

lfittl (approved):

components/MonitoringUserSetupInstructions.tsx
uhoh-itsmaciek: Yeah. Unfortunately, I'm not sure if there's a good way to detect that, but I have not seen any support requests around this (and it should break log-based EXPLAIN if we don't strip the trailing semicolon correctly.). We could announce it in e.g., the newsletter, or the next collector release notes, but given the limited impact, I think we can just make the change silently.

PR #103, opened 2022-08-02
Enterprise: Rework documentation, consistently use "Enterprise Server"

lfittl:
This clarifies that object storage is useful beyond just Log Insights,
adds a dedicated page for installing the collector separately, and
adds a new page that describes a scaled out reference architecture.

Additionally we use "Enterprise Server" going forward, instead of
"Enterprise Edition".

### DIFF
--- a/enterprise/architecture.mdx
+++ b/enterprise/architecture.mdx
@@ -1,0 +1,11 @@
+---
+title: 'Enterprise Server: Reference Architecture'
+backlink_href: /docs/enterprise
+backlink_title: 'pganalyze Enterprise Server'
+---
+
+Here you can see the reference architecture for a scaled out version of pganalyze Enterprise Server. This diagram also illustrates how data gets sent between different components.
+
+This is focused on a deployment for monitoring Amazon RDS or Aurora, [deployed using Amazon EKS](/docs/enterprise/setup/kubernetes_amazon_eks) and intended for larger setups that [run the collector separately](/docs/enterprise/setup/separate-collector-install), with a separation between the central AWS account and the application AWS accounts.
+
+![Reference architecture diagram](architecture.svg)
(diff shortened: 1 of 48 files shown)


enterprise/architecture.mdx
```diff
@@ -0,0 +1,11 @@
...
+title: 'Enterprise Server: Reference Architecture'
+backlink_href: /docs/enterprise
+backlink_title: 'pganalyze Enterprise Server'
+---
+
+Here you can see the reference architecture for a scaled out version of pganalyze Enterprise Server. This diagram also illustrates how data gets sent between different components.
+
+This is focused on a deployment for monitoring Amazon RDS or Aurora, [deployed using Amazon EKS](/docs/enterprise/setup/kubernetes_amazon_eks) and intended for larger setups that [run the collector separately](/docs/enterprise/setup/separate-collector-install), with a separation between the central AWS account and the application AWS accounts.
+
+![Reference architecture diagram](architecture.svg)
```
uhoh-itsmaciek, line 11: This is great. I think at some point, we should also add a scaled-down version (i.e., internal collector, app running on-prem), but this is a better starting point.

open_source_components.mdx
```diff
@@ -4,19 +4,6 @@ backlink_href: /docs
...
-## pgdatagraph
-
-![](https://raw.githubusercontent.com/pganalyze/pgdatagraph/master/screenshot.png)
-
-[pgdatagraph](https://github.com/pganalyze/pgdatagraph) is an interactive widget
-for graphing time-series data - based on [D3](http://d3js.org/) and
-[Rickshaw](http://code.shutterstock.com/rickshaw/).
-
-It's used throughout pganalyze for all the graphs you see. Its goal is to enable
-exploration and interaction with the data.
```
uhoh-itsmaciek, line 16: Ha, good catch.

uhoh-itsmaciek (approved):

enterprise/architecture.mdx
lfittl: Yeah, agreed - this was already done so I thought why not share it with more people - agreed more follow-on work would be good.

PR #106, opened 2022-08-11
Update SSO docs to reflect pending changes

msakrejda:
- Reflow SAML docs to make them easier to edit
- Note that you can now specify IdP fields directly
- Update SSO docs to reflect pending changes with migrating accounts


accounts/sso/saml2.mdx
```diff
@@ -50,13 +50,13 @@ To tell pganalyze which IdP you trust, you need to retrieve the metadata URL
...
 
-Note that this is the only method of configuration supported, at this point you
-cannot specify Service Provider (SP) settings directly. Please reach out to
-[pganalyze Support](mailto:support@pganalyze.com) for alternate setup methods.
+Alternately, you can specify Service Provider (SP) settings directly:
 
-Once you submit the Metadata URL, pganalyze will verify the configuration, and
+![Screenshot of pganalyze SAML integration form for the IdP metadata details](sso_idp_metadata_form_details.png)
+
+Once you submit the metadata, pganalyze will verify the configuration, and
```
lfittl, line 59: Technically in the case of providing metadata directly (instead of through the URL) we don't do any verification, we just remember the settings. When you attempt to sign in that's the time we really know whether its working or not.

lfittl (approved):
:+1: (with one minor note)

accounts/sso/saml2.mdx
msakrejda: Do we do verification through the metadata URL? I only see that going through OneLogin::RubySaml::IdpMetadataParser, but that doesn't seem to validate much, looking at the code. Maybe we should add validation in both cases? E.g., we could at least call `OneLogin::RubySaml::Utils.is_cert_expired` to check the cert is current and well-formed?

lfittl: @msakrejda Yeah, we could certainly do that - that'd also make sure the certificate itself wasn't pasted in incorrectly.

For the existing logic for URL access I'd argue that does actually do validation in the sense that it ensures the URL is accessible (since it downloads the data), which would fail if you provided the incorrect URL.

msakrejda: Sure, makes sense. And I guess once we validate that the metadata URL itself is accessible, the metadata loaded from there is reasonably likely to be correct (versus something just copy-pasted).

msakrejda: See https://github.com/pganalyze/pganalyze/pull/2242

PR #114, opened 2022-11-08
Add collector version note to stats column helper troubleshooting

msakrejda:


install/troubleshooting/column_stats_helper.mdx
```diff
@@ -12,12 +12,21 @@ You may get the following warning when running a collector test:
...
+```
+
 This warning indicates that column statistics for the specific database are not being reported by the collector.
 Column statistics are used by pganalyze to improve Index Advisor recommendations. Index Advisor will still work
 without the helper functions, but index recommendations may be less accurate.
 
-To resolve this, find the name for the database role that's configured as the monitoring user
-in your collector installation (we'll assume "pganalyze" here).
+To resolve this, first, make sure you are running at least collector version 0.41.0 (and ideally the latest version).
+You can see which version you are currently running by checking the Debug Info panel on the Settings page for your
```
keiko713, line 27: I wonder if it might be worth mentioning that it's under "Server Settings" tab. Or, change it to something like "Settings > Server Settings > Debug Info" (though maybe we're not using this pattern to explain the navigation).
I was able to find the Debug Info panel but did get confused a tiny bit. Not a strong opinion, the current wording works too.

keiko713 (approved):
Overall looking good 👍

install/troubleshooting/column_stats_helper.mdx
msakrejda: Good point, I'll mention the tab. We don't use the ">" pattern so I'll avoid introducing that here. I think with just the ASCII characters it looks a little awkward, but maybe we should think about introducing something like that that we could use with FontAwesome icons.

PR #119, opened 2022-11-24
Update blocking query issue details key name for autovacuum wraparound

keiko713:
What title says. It'll be changed in the main app side, so let's change this in the docs too.


lfittl (approved):
:+1:

PR #124, opened 2022-12-03
Drop postgresql-contrib from Debian-based platform instructions

msakrejda:
This package is no longer needed--the core extensions (specifically,
pg_stat_statements in our case) come bundled with Postgres.


lfittl (approved):
:+1:

PR #125, opened 2022-12-03
Add Table of Contents (TOC) metadata files and remove index contents

lfittl:
Instead of using the index as a way to find pages, this is a part of a change that introduces a side navigation that contains the ToC and is available on every page.


index.mdx
```diff
@@ -2,129 +2,4 @@
...
-  - [Tuning checkpoint intervals to reduce I/O spikes](/docs/guides/tuning-checkpoint-intervals)
-  - [Adjusting work_mem based on temporary file creation](/docs/guides/adjusting-work-mem)
-  - [Exporting query statistics using the pganalyze API](/docs/guides/exporting-query-statistics)
-  - [Monitoring Postgres locks using Log Insights](/docs/guides/monitoring-postgres-locks-using-log-insights)
-*/}</>
-
----
-
-* [Open-Source Components](/docs/open_source_components)
+Please select a navigation item on the left.
```
lfittl, line 5: Note this is intended to be merged at the same time as the website change that introduces the side navigation as a replacement.

style.module.scss
```diff
@@ -1,3 +1,82 @@
...
+  margin-top: 15px;
+}
+.overviewInstallChoice + h2, .overviewFeatureChoice + h2 {
+  margin-top: 32px;
+}
+.overviewInstallChoiceStep, .overviewFeatureChoiceStep {
+  width: 90%;
+  display: flex;
+  justify-content: space-between;
+  align-items: left;
```
msakrejda, line 11: This is [not a valid value for this property](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#syntax). I'd suggest an alternative, but I'm not sure I follow the intent here, since "left" can't apply to the cross axis with the default "row" `flex-direction` and `justify-contents` is set above for content alignment.

style.module.scss
```diff
@@ -1,3 +1,82 @@
...
+  .overviewInstallChoiceStep img {
+    order: 0;
+  }
+  .overviewInstallChoice, .overviewFeatureChoice {
+    display: flex;
+    margin-bottom: 10px;
+  }
+  .overviewFeatureChoiceStep {
+    justify-content: space-between;
+    align-items: left;
```
msakrejda, line 68: Same issue as above.

style.module.scss
```diff
@@ -1,3 +1,82 @@
...
+  }
+  .overviewInstallChoiceStep, .overviewFeatureChoiceStep {
+    flex-grow: 1;
+    width: 33%;
+    height: 80px;
```
msakrejda, line 73: Setting a fixed height here means text can overflow out of the container:

image

If we want a bunch of aligned tiles, maybe `grid` would be a better mechanism? We should be able to do something like
```css
.overviewInstallChoiceStep, .overviewFeatureChoiceStep {
  display: grid;
  grid-template-columns: 1fr 1fr;
  gap: 20px;
}
```
to simplify this styling. (We can override `grid-template-columns` to one column for smaller viewports.) I can explore this if you'd like.

style.module.scss
```diff
@@ -1,3 +1,82 @@
...
+  }
+
+  img, strong {
+    align-self: center;
+    text-align: center;
+  }
+}
+
+.overviewInstallChoiceStep {
+  height: 50px;
```
msakrejda, line 52: Same as below, a fixed height may cause overflow here.

msakrejda (approved):

style.module.scss
lfittl: Implemented, per your suggestion - thanks for the feedback!

PR #126, opened 2022-12-10
Document `always_collect_system_data` field

msepga:
Added in pganalyze/collector#341. Documents the option under the 'Self-managed servers' section on the collector settings page.


lfittl (approved):
:+1: Thanks for adding this!

PR #134, opened 2023-02-28
Collector settings: Use "=" instead of ":" to separate key values

lfittl:
This is how we typically show other config examples, as well as the example config file that's installed with the packages.

In passing, clarify that the file is located at /etc/pganalyze-collector.conf (not just "pganalyze-collector.conf") which has been a confusion for some users, and in 99% of the cases this is the correct location.


keiko713 (approved):
📜

msakrejda (approved):

PR #135, opened 2023-03-01
Update monitoring user setup instructions

msakrejda:
- drop 9.6 instructions (since we've dropped 9.6 support)
 - drop the minPostgres flag to the MonitoringUserSetupInstructions
   component since instructions are now uniform
 - use the pre-pg_monitor instructions for pg_stat_statements_reset
   on current versions, since this is not actually included in
   pg_monitor permissions

Regarding the last point, our docs implied that pg_monitor grants the
ability to call pg_stat_statements_reset(), but I'm not sure when that
would have been in effect. The role was new as of 10, losing that
ability doesn't seem to be mentioned in the 11.0, 12.0, or 13.0
release notes, and it doesn't seem to work in the 10, 14, or 15 that I
have installed locally.

### DIFF
--- a/components/MonitoringUserSetupInstructions.tsx
+++ b/components/MonitoringUserSetupInstructions.tsx
@@ -1,41 +1,21 @@
 import React from "react";
 
 import { useCodeBlock } from "./CodeBlock";
-import TabPanel, { TabItem } from './TabPanel'
 
 type Props = {
-  minPostgres: number;
   password: string;
   adminUsername?: string;
   noPgMonitor?: boolean;
 };
 
 const MonitoringUserSetupInstructions: React.FunctionComponent<Props> = ({
-  minPostgres = 90600,
   password = 'mypassword',
   adminUsername,
   noPgMonitor
 }) => {
-  type SetupOpt = [
-    id: string,
-    version: number,
-    label: string,
-    instructions: React.ComponentType<{password: string, noPgMonitor?: boolean, adminUsername?: string}>
-  ];
-  const opts: SetupOpt[] = [
-    [ 'monitoring_user_10', 100000, 'PostgreSQL 10+', MonitoringUser10 ],
-    [ 'monitoring_user_96', 90600, 'PostgreSQL 9.6', MonitoringUser96 ],
-  ].filter(([_id, version, _label ]) => minPostgres <= version) as SetupOpt[];
-
-  const tabs = opts.map<TabItem>(opt => [opt[0], opt[2]])
   return (
     <>
-      <TabPanel items={tabs}>
-        {(idx: number) => {
-          const ActiveTab = opts[idx][3];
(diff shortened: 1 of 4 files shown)


keiko713:
> use the pre-pg_monitor instructions for pg_stat_statements_reset

Did you actually make this change? Which part is it?

msakrejda:
@keiko713 yeah, sorry, that part is pretty subtle: It's just making [this part of the instructions](https://github.com/pganalyze/pganalyze-docs/pull/135/files#diff-29b4a130316538bca91876ddac10f2adc7140298e786bd334b0cc8aa71a365c0R155-R157), which was previously 9.6-only, visible in the general instructions.

keiko713 (approved):
Unsure about the `pg_stat_statements_reset()` ability either, though your test sounds pretty affirmative so I think we can go with this way.

lfittl:
> our docs implied that pg_monitor grants the
> ability to call pg_stat_statements_reset(), but I'm not sure when that
> would have been in effect.

I think we wrote our docs that way initially because pg_monitor initially allowed a reset, and then it was removed in https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=edb9797660541b217d23ae7c02b96b496d34fec4 (which was a commit to Postgres 12, that was backpatched to 11 and 10)

I think the way this is done here makes sense, i.e. to always describe adding that helper.

components/MonitoringUserSetupInstructions.tsx
```diff
@@ -172,52 +152,8 @@ $$
...
-
-CREATE OR REPLACE FUNCTION pganalyze.get_stat_replication() RETURNS SETOF pg_stat_replication AS
-$$
-  /* pganalyze-collector */ SELECT * FROM pg_catalog.pg_stat_replication;
-$$ LANGUAGE sql VOLATILE SECURITY DEFINER;
-
-CREATE OR REPLACE FUNCTION pganalyze.get_stat_progress_vacuum() RETURNS SETOF pg_stat_progress_vacuum AS
-$$
-  /* pganalyze-collector */ SELECT * FROM pg_catalog.pg_stat_progress_vacuum;
-$$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}</CodeBlock>
```
lfittl, line 218: FWIW, I think it may still be useful to describe these somewhere in our docs (since not all platforms allow pg_monitor access), but it doesn't have to be in this component (which is used in the setup guide for supported platforms), so makes sense to remove for now.

lfittl (approved):

components/MonitoringUserSetupInstructions.tsx
msakrejda: Makes sense. If nothing else, we could have a troubleshooting section on monitoring without `pg_monitor`.

PR #137, opened 2023-03-02
Enterprise release notes for 2023.03.0

lfittl:


enterprise/releases/2023-03-0.mdx
```diff
@@ -0,0 +1,122 @@
...
+  - New per-table graph that shows the age of the oldest unfrozen XID on a table (to help understand anti-wraparound vacuums better)
+  - Extend VACUUM data retention time to 35 days (from 7 days)
+  - Additionally this release includes preparation for new autovacuum metrics and tuning recommendations, coming in a future Enterprise Server release
+- Other improvements
+  - Show Amazon RDS / Aurora instance tags on servers in pganalyze
+    - For instance with tags applied in AWS (e.g. to indicate which application team owns the instance),
+      pganalyze will now automatically show these tags in the server overview list
+  - Add `DISABLE_DIRECT_SIGNUP` setting to prevent uninvited users from registering a user account
+    - In case you are hosting your pganalyze instance on a publicly accessible host, we recommend enabling this new setting as an extra measure to prevent unexpected user registrations
+    - Note that this setting is not compatible with the `DEFAULT_ORG_ROLE` setting (which when enabled automatically adds new users to the initial organization in pganalyze)
```
msakrejda, line 74: If users do set both, is the behavior well defined? Should we say what happens?

enterprise/releases/2023-03-0.mdx
```diff
@@ -0,0 +1,122 @@
...
+  - Add support for AWS IMDSv2 instance metadata protocol to retrieve credentials for accessing object storage
+  - Table columns page: Show NULL % and avg. width (requires [pganalyze.get_column_stats](/docs/install/amazon_rds/01_configure_rds_instance) helper)
```
msakrejda, line 76: I think this is pointing to the wrong page?
```suggestion
  - Table columns page: Show NULL % and avg. width (requires [pganalyze.get_column_stats](/docs/install/amazon_rds/02_create_monitoring_user) helper)
```
We should probably have standalone troubleshooting pages for these, but that's a separate problem.

msakrejda (approved):
Some minor comments, but looks great.

enterprise/releases/2023-03-0.mdx
lfittl: Yeah, good point - I moved the explanation to the settings page instead.

PR #142, opened 2023-03-11
Document new and updated collector settings

msakrejda:
- disable_citus_schema_stats
 - db_use_iam_auth
 - new rds-ca-global option for db_sslrootcert

### DIFF
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -141,7 +141,7 @@ helper functions are expected to be defined, and the one we connect to for serve
     <tr>
       <td>db_sslrootcert (<code>DB_SSLROOTCERT</code>)</td>
       <td>system certificate store</td>
-      <td>Path to SSL certificate authority (CA) certificate(s) to use to verify the server's certificate, or <code>rds-ca-global</code> to use the built-in AWS RDS certificates</td>
+      <td>Path to SSL certificate authority (CA) certificate(s) to use to verify the server's certificate, or <code>rds-ca-global</code> to use the built-in Amazon RDS global CA certificate set</td>
     </tr>
     <tr>
       <td>db_sslrootcert_contents (<code>DB_SSLROOTCERT_CONTENTS</code>)</td>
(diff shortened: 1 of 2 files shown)


collector/settings.mdx
```diff
@@ -305,6 +305,16 @@ Note that the `aws_endpoint_*` settings are only relevant if you are using custo
       <td>If specified, and api_system_scope (see below) is not specified, this is prepended to the auto-generated system scope (optional, can be used
       to, e.g., differentiate staging from production)</td>
     </tr>
+    <tr>
+      <td>db_use_iam_auth (<code>DB_USE_IAM_AUTH</code>)</td>
```
msakrejda, line 309: I noticed this is the only AWS-specific setting that does not start with `aws_`. Do we want to rename it prior to documenting it?

lfittl: That was intentional, Google also calls their implementation "IAM authentication", and so it seemed like we could use this as a more general term. See https://github.com/pganalyze/collector/pull/368#discussion_r1087387656

collector/settings.mdx
```diff
@@ -537,6 +547,15 @@ talking to pganalyze support.
       <td>Experimental: name of docker container to collect logs from using <code>docker logs -t</code>.
       This requires that the collector runs on the Docker host.</td>
     </tr>
+    <tr>
+      <td>disable_citus_schema_stats (<code>DISABLE_CITUS_SCHEMA_STATS</code>)</td>
+      <td>[none]</td>
+      <td>
+        If using the Citus extension in your database, turn off the collection of statistics for distributed indexes
+        and tables. For very large schemas, this collection can be a performance problem. When using this option it's
```
lfittl, line 555: Maybe "can be a performance problem" => "can error out due to timeouts or locks"

I think our statement timeout makes sure this isn't an actual performance problem, but if you have many tables its likely to either run into timeout issues (i.e. it takes too long), or locking problems because of how this is implemented in Citus.

collector/settings.mdx
```diff
@@ -141,7 +141,7 @@ helper functions are expected to be defined, and the one we connect to for serve
     <tr>
       <td>db_sslrootcert (<code>DB_SSLROOTCERT</code>)</td>
       <td>system certificate store</td>
-      <td>Path to SSL certificate authority (CA) certificate(s) to use to verify the server's certificate, or one of <code>rds-ca-2015-root</code> or <code>rds-ca-2019-root</code> to use the built-in AWS RDS certificates</td>
+      <td>Path to SSL certificate authority (CA) certificate(s) to use to verify the server's certificate, or <code>rds-ca-global</code> to use the built-in AWS RDS certificates</td>
```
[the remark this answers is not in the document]
lfittl: Also, we should adjust the RDS setup instructions to use "rds-ca-global" instead of "rds-ca-2019-root" (since the 2019 CA is included in the global bundle, its safe to just replace it everywhere)

lfittl (approved):

collector/settings.mdx
msakrejda: Okay. I think it maybe would have made more sense to have parallel `aws_use_iam_auth`, `gcp_use_iam_auth`, etc., but this is reasonable. I imagine we still want to keep it in the AWS section, though, at least for now?

msakrejda: Oh, just saw your other comment. I'll keep it here.

collector/settings.mdx
```diff
@@ -305,6 +305,16 @@ Note that the `aws_endpoint_*` settings are only relevant if you are using custo
...
       to, e.g., differentiate staging from production)</td>
     </tr>
+    <tr>
+      <td>db_use_iam_auth (<code>DB_USE_IAM_AUTH</code>)</td>
+      <td>[none]</td>
+      <td>
+        Fetches a short-lived token for logging into the database instance from the AWS API, instead of using a hardcoded password
+        in the collector configuration file. To use this setting, IAM authentication needs to be enabled on the database instance
+        / cluster, and the pganalyze IAM policy needs to <a href="https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html">cover the "rds-db:connect" privilege</a> for
+        the pganalyze user.
```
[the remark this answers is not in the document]
msakrejda: Makes sense. I was mostly copying from the release notes. I'll adjust.

PR #144, opened 2023-04-01
Improve Enterprise Server setup docs for better clarity

lfittl:
- Use /etc/pganalyze.env as the recommended config file location
- List both Docker and podman as supported container runtimes
- Remove MAILER_URL from initial setup steps
- Remove outdated database dashboard screenshot
- Clarify that object storage setup is recommended for most setups
- MinIO: Provide copy and paste steps for creation of access/secret keys
- MinIO: Clarify how to set MINIO_ENDPOINT
- MinIO: Fix retention time of logs bucket to be 7 days, not 35 days
- add missing ToC entries for object storage setup docs pages


enterprise/setup/index.mdx
```diff
@@ -48,29 +51,32 @@ docker pull quay.io/pganalyze/enterprise:[version]
...
+LICENSE_KEY=REPLACE_ME
 ```
 
-If you'd like to use the Log Insights or Automated EXPLAIN features, or want to run the collector separately from the Enterprise Server container, you also need to pass the [object storage configuration](/docs/enterprise/setup/object-storage) here.
+You can find the full explanation of all variables in the [settings documentation](/docs/enterprise/settings), for example the `MAILER_URL` setting which is required to activate email alerts.
 
-You can find the full explanation of all variables in the [settings documentation](/docs/enterprise/settings).
+### Step 3: Set up object storage
 
+If you'd like to use the Log Insights or Automated EXPLAIN features, or want to run the collector separately from the Enterprise Server container, we now recommend doing the [object storage configuration](/docs/enterprise/setup/object-storage).
```
keiko713, line 71: [nitpick] maybe you are saying "we now recommend" here because of the historical reason? (e.g. it was not the recommendation previously)
Lots of folks who will see here don't have the historical context so I think you can drop "now" there.

enterprise/setup/index.mdx
```diff
@@ -48,29 +51,32 @@ docker pull quay.io/pganalyze/enterprise:[version]
...
 
-### Step 3: Initialize Database & Verify Installation
+### Step 4: Initialize database & verify installation
 
-First, run the Enterprise self check, to ensure the configuration is correct:
+Next, run the Enterprise self check, to ensure the configuration is correct:
```
keiko713, line 75: [nitpick] There is no "First" in the new version of doc, yet the following step 5 section also starts with "Next". I guess you can say "Next" followed by "Next", though it might feel a little awkward. Alternatively, we can remove these "Next" and change this sentence to "Run the Enterprise self check to ensure the configuration is correct" (then step 5 will be "Create the initial user").

enterprise/setup/kubernetes_amazon_eks.mdx
```diff
@@ -416,11 +416,15 @@ Once data is coming in successfully, the "Debug Info" looks like this:
 
 ![](debug_info_with_data.png)
 
-Be aware that some graphs need at least a few hours worth of data and might not work properly before. It is therefore recommended that you wait for up to 24 hours before evaluating the system more closely.
+**Important:** Be aware that some graphs need at least a few hours worth of data and might not work properly before.
```
keiko713, line 419: (just an idea, no need to address in this PR) we could use callout box for things like this

keiko713 (approved):
left several nitpicks (no need to address), but looks good to me 👍

enterprise/setup/index.mdx
lfittl: Hmm, yeah, this was meant in reference to "you're working through this guide, and *now* as part of this step you can do this". But I think the now is unnecessary and I'll remove it to avoid confusion.

enterprise/setup/kubernetes_amazon_eks.mdx
lfittl: Yeah, that would be good - I don't think we have a style for this in the docs right now, so will skip for now, but would be good to add (I think when we add it, grepping for "Important" should find the relevant occurrences)

PR #148, opened 2023-04-22
Add instruction for sidekiq prometheus monitoring

keiko713:
This was introduced with enterprise-v2023.03.0 release.
Documenting here so that the Enterprise customers could take advantage of it potentially.

### DIFF
--- a/enterprise/settings.mdx
+++ b/enterprise/settings.mdx
@@ -38,7 +38,7 @@ MAILER_FROM=pganalyze@example.com
   - `modify_all` - assigns the role named "View & Modify (All Servers)"
   - `view_all` - assigns the role named "View (All Servers)"
 
-**Important:** If you are hosting your pganalyze Enterprise Server installation on a publicly available interface (i.e. without explicit access control at the network level), we recommend setting `DISABLE_DIRECT_SIGNUP` to `1`. 
+**Important:** If you are hosting your pganalyze Enterprise Server installation on a publicly available interface (i.e. without explicit access control at the network level), we recommend setting `DISABLE_DIRECT_SIGNUP` to `1`.
 
 Note that `DISABLE_DIRECT_SIGNUP` is not compatible with the `DEFAULT_ORG_ROLE` setting: If `DISABLE_DIRECT_SIGNUP` is enabled, and `DEFAULT_ORG_ROLE` is not `none`, it will be the same as setting it to `none`, since direct registrations are not possible.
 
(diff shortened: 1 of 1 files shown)


enterprise/settings.mdx
```diff
@@ -128,3 +128,23 @@ DEFAULT_ORG_ROLE=view_all
 ```
 
 Note you can run an [Enterprise self-check](/docs/enterprise/troubleshooting) to verify the connection and lookup LDAP information, but some configuration errors are only visible once logging in.
+
+## Sidekiq monitoring using Prometheus
+
+The metrics HTTP endpoint `/metrics` is available for the Prometheus scraping. Currently, it provides Sidekiq metrics using [sidekiq-prometheus-exporter gem](https://github.com/Strech/sidekiq-prometheus-exporter).
```
lfittl, line 134: From a grammatical perspective, I think the gem should be prefixed with "the". Separately, I think there should be no "the" prefixing Prometheus.

Maybe we can rewrite like this:

```suggestion
The metrics HTTP endpoint `/metrics` is available for Prometheus scraping. Currently, it provides worker queue metrics using the [sidekiq-prometheus-exporter gem](https://github.com/Strech/sidekiq-prometheus-exporter).
```

enterprise/settings.mdx
```diff
@@ -128,3 +128,23 @@ DEFAULT_ORG_ROLE=view_all
 ```
 
 Note you can run an [Enterprise self-check](/docs/enterprise/troubleshooting) to verify the connection and lookup LDAP information, but some configuration errors are only visible once logging in.
+
+## Sidekiq monitoring using Prometheus
```
lfittl, line 132: Maybe "Worker queue monitoring using Prometheus"?

(the fact that we're using Sidekiq is an implementation detail in a sense)

enterprise/settings.mdx
```diff
@@ -128,3 +128,23 @@ DEFAULT_ORG_ROLE=view_all
...
+
+The metrics HTTP endpoint `/metrics` is available for the Prometheus scraping. Currently, it provides Sidekiq metrics using [sidekiq-prometheus-exporter gem](https://github.com/Strech/sidekiq-prometheus-exporter).
+
+The endpoint is protected with a bearer auth, the token is specified with the `SIDEKIQ_PROMETHEUS_TOKEN` environment variable. The endpoint is only available when the environment variable is set.
+If you have a Prometheus setup already, you can add the hostname and port of your web server as an additional target to the prometheus config file to scrape.
```
lfittl, line 137: For consistency, we should probably uppercase Prometheus in the later part of the sentence too:

```suggestion
If you have a Prometheus setup already, you can add the hostname and port of your web server as an additional target to the Prometheus config file to scrape.
```

lfittl (approved):
Thanks for adding this!

Some minor notes, but looks good overall. 🚢

enterprise/settings.mdx
keiko713:
```suggestion
## Worker queue monitoring using Prometheus
```

PR #149, opened 2023-04-26
Add VACUUM: Inefficient Index Phase check

msakrejda:


components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const threshold = config.settings["threshold_count"] as number;
+  const runPluralized = threshold === 1 ? 'run' : 'runs';
+  return (
+    <>
+      <p>
+        Detects when autovacuum runs in this server are forced to perform inefficient
+        multiple index scan phases due to limited configured memory, and creates an
+        issue with severity "warning". Triggers when at least <code>{threshold}</code>
+        autovacuum {runPluralized} in the last 24h have had multiple index scan phases.
+        Resolves automatically once fewer than <code>{threshold}</code> autovacuum {runPluralized}
```
keiko713, line 17: Uhmm any idea why this is happening? Or it's the old version screenshot?

image

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
```diff
@@ -0,0 +1,76 @@
...
+        Autovacuum on affected tables may take significantly longer. VACUUM processes tables
+        and their indexes in phases. To avoid impacting your primary workload, autovacuum
+        is limited to using a certain amount of memory. If this is not enough to process
+        all dead rows, the index scan phase will need to be repeated (possibly several times).
+        This is inefficient in terms of CPU and I/O.
+      </p>
+      <h4>Common Causes</h4>
+      <ul>
+        <li>
+          <h5><code>autovacuum_work_mem</code> set too low</h5>
```
keiko713, line 44: I have a slightly mixed feeling about having `<code>` in `<h5>` (not strong feeling). What do you think about saying something like "Autovacuum work memory set too low"?

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <p>
+        Detects when autovacuum runs in this server are forced to perform inefficient
+        multiple index scan phases due to limited configured memory, and creates an
+        issue with severity "warning". Triggers when at least <code>{threshold}</code>
+        autovacuum {runPluralized} in the last 24h have had multiple index scan phases.
+        Resolves automatically once fewer than <code>{threshold}</code> autovacuum {runPluralized}
+        in the last 24h have had multiple index scan phases.
+      </p>
+      <p>
+        Ignores situations where configured autovacuum mmemory is already set to the
```
keiko713, line 21:
```suggestion
        Ignores situations where configured autovacuum memory is already set to the
```

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
```diff
@@ -0,0 +1,76 @@
...
+            If VACUUM is not occurring often enough, many dead rows can accumulate between
+            runs, and will require more memory to clean up.
+          </p>
+        </li>
+      </ul>
+      <h4>Solution</h4>
+      <p>
+        Increasing <SettingLink setting="autovacuum_work_mem" /> can allow autovacuum
+        to proceed more efficiently. It's hard to predict how much memory may be needed,
+        but doubling the value and monitoring the impact is a reasonable approach.
```
keiko713, line 63: It would be helpful to add the description (or query) of how to double the value.

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
msakrejda: Ah, good catch. This needs a `{" "}` due to [whitespace handling in JSX](https://stackoverflow.com/a/32690647).

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
msakrejda: I thought about that too, but it looks okay to me when rendered. I don't mind changing it, though.

keiko713: I don't mind leaving it as is too :P well let's keep it as is

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
msakrejda: Sure, makes sense. The `work_mem` check says 
```jsx
<>
Update the setting to the recommended value by using{" "}
<PGDocsLink path="/sql-altersystem.html">ALTER SYSTEM</PGDocsLink> or
modifying the parameters in your cloud provider portal.
</>
```
I think I can just copy that here.

msakrejda:
I think I've addressed all review feedback, and I added an explanation of the `maintenance_work_mem` fallback. This is ready for review.

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
keiko713: @msakrejda looks like this is not fixed yet

keiko713 (approved):
Still one typo but looks good otherwise!

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
```diff
@@ -0,0 +1,86 @@
...
+  config,
+}) => {
+  const threshold = config.settings["threshold_count"] as number;
+  const runPluralized = threshold === 1 ? 'run' : 'runs';
+  return (
+    <>
+      <p>
+        Detects when autovacuum runs in this server are forced to perform inefficient
+        multiple index scan phases due to limited configured memory, and creates an
+        issue with severity "warning". Triggers when at least <code>{threshold}</code>
```
keiko713, line 16: Actually this also needs `{" "}` I believe

components/CheckDocumentation/vacuum/InefficientIndexPhase.tsx
```diff
@@ -0,0 +1,86 @@
...
+          </p>
+        </li>
+      </ul>
+      <h4>Solution</h4>
+      <p>
+        Increasing <SettingLink setting="autovacuum_work_mem" /> can allow autovacuum
+        to proceed more efficiently. It's hard to predict how much memory may be needed,
+        but doubling the value (up to the maximum of <code>1GB</code>) and monitoring the
+        impact is a reasonable approach. If <code>autovacuum_work_mem</code> is set to
+        <code>-1</code>, it will fall back to the current value of <SettingLink setting="maintenance_work_mem" />.
```
keiko713, line 70: Uhmm I guess I'm a space cop. This also looks like needing `{" "}` or something before `<code>-1</code>`

PR #151, opened 2023-05-15
API docs: Fix getIssues to avoid deprecated fields, other minor edits

lfittl:


api/queries/getIssues.mdx
```diff
@@ -8,14 +8,15 @@ backlink_title: 'pganalyze GraphQL API Queries'
...
+them to a third-party system not directly supported by pganalyze's
+[alerting integrations](/docs/checks/#alerting).
 
 ## API
 
 Arguments for `getIssues`:
 
-* **`databaseId` (string)**<br/>Database ID for which to retrieve issues for
-* **`serverId` (string)**<br/>Server ID for which to retrieve issues for (either this or `databaseID` needs to be specified)
+* **`serverId` (string)**<br/>Server ID for which to retrieve issues for
```
keiko713, line 18: Just curious: is there any way to mention that this is a "human_id", not UUID? The example kinda demonstrates it but I wonder if further clarification would be helpful

keiko713 (approved):
🪧

api/queries/getIssues.mdx
lfittl: Yeah, good question - I think we could probably benefit from a "What's a Server ID" section that describes this in more detail (especially when working with API responses that may contain the raw ID), but given that we mainly use the human IDs in all customer-facing contexts, I think its okay to omit this detail here.

PR #154, opened 2023-05-25
Add VACUUM: Bloat - VACUUM Blocked By Xmin Horizon check

keiko713:
Add a check guidance for "VACUUM: Bloat - VACUUM Blocked By Xmin Horizon" check. The below shows all 5 causes, but in app, we will only show the specific cause that triggered a check.

In the check configuration, there is a blank "hours ago" but it's just due to some browser cache problem, it's implemented okay.

image

### DIFF
--- a/components/CheckDocumentation/vacuum/XminHorizon.tsx
+++ b/components/CheckDocumentation/vacuum/XminHorizon.tsx
@@ -63,38 +63,28 @@ const XminHorizonGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
       </p>
       <h4>Common Causes and Solutions</h4>
       <ul>
-        <li>
-          <GuidanceByBackend
-            inApp={inApp}
-            xmin={byBackend && byBackend["xmin"]}
-          />
-        </li>
-        <li>
-          <GuidanceByReplicationSlot
-            inApp={inApp}
-            xmin={byReplicationSlot && byReplicationSlot["xmin"]}
-            serverReplicationUrl={serverReplicationUrl}
-          />
-        </li>
-        <li>
-          <GuidanceByReplicationSlotCatalog
-            inApp={inApp}
-            xmin={byReplicationSlotCatalog && byReplicationSlotCatalog["xmin"]}
-            serverReplicationUrl={serverReplicationUrl}
-          />
-        </li>
-        <li>
-          <GuidanceByStandby
-            inApp={inApp}
-            xmin={byStandby && byStandby["xmin"]}
-          />
-        </li>
-        <li>
-          <GuidanceByPreparedXact
-            inApp={inApp}
-            xmin={byPreparedXact && byPreparedXact["xmin"]}
-          />
-        </li>
+        <GuidanceByBackend
+          inApp={inApp}
+          xmin={byBackend && byBackend["xmin"]}
+        />
+        <GuidanceByReplicationSlot
+          inApp={inApp}
+          xmin={byReplicationSlot && byReplicationSlot["xmin"]}
+          serverReplicationUrl={serverReplicationUrl}
+        />
+        <GuidanceByReplicationSlotCatalog
+          inApp={inApp}
+          xmin={byReplicationSlotCatalog && byReplicationSlotCatalog["xmin"]}
+          serverReplicationUrl={serverReplicationUrl}
+        />
+        <GuidanceByStandby
+          inApp={inApp}
+          xmin={byStandby && byStandby["xmin"]}
+        />
+        <GuidanceByPreparedXact
+          inApp={inApp}
+          xmin={byPreparedXact && byPreparedXact["xmin"]}
+        />
       </ul>
     </div>
   );
@@ -110,7 +100,7 @@ const GuidanceByBackend: React.FunctionComponent<{
 
   const CodeBlock = useCodeBlock();
   return (
-    <>
+    <li>
       <h5>Long-running transactions</h5>
       <p>
         Long-running transactions may still need to access rows that could
@@ -142,7 +132,7 @@ const GuidanceByBackend: React.FunctionComponent<{
                 SELECT pg_terminate_backend('<query_pid>');`}
         />
       </CodeBlock>
-    </>
+    </li>
   );
 };
 
@@ -158,7 +148,7 @@ const GuidanceByReplicationSlot: React.FunctionComponent<{
   const Link = useSmartAnchor();
   const CodeBlock = useCodeBlock();
   return (
-    <>
+    <li>
       <h5>Lagging or stale physical replication slots</h5>
       <p>
         With physical streaming replication with{" "}
@@ -185,7 +175,7 @@ const GuidanceByReplicationSlot: React.FunctionComponent<{
       <CodeBlock>
         <SQL sql={`SELECT pg_drop_replication_slot('<slot_name>');`} />
       </CodeBlock>
-    </>
+    </li>
   );
 };
 
@@ -201,7 +191,7 @@ const GuidanceByReplicationSlotCatalog: React.FunctionComponent<{
   const CodeBlock = useCodeBlock();
   const Link = useSmartAnchor();
   return (
-    <>
+    <li>
       <h5>Lagging or stale logical replication slots</h5>
       <p>
         With logical replication slots, replication can also get stale when DDL
@@ -232,7 +222,7 @@ const GuidanceByReplicationSlotCatalog: React.FunctionComponent<{
       <CodeBlock>
         <SQL sql={`SELECT pg_drop_replication_slot('<slot_name>');`} />
       </CodeBlock>
-    </>
+    </li>
   );
 };
 
@@ -246,7 +236,7 @@ const GuidanceByStandby: React.FunctionComponent<{
 
   const CodeBlock = useCodeBlock();
   return (
-    <>
+    <li>
       <h5>Long-running queries on standbys</h5>
       <p>
         When <code>hot_standby_feedback</code> is on with physical streaming
@@ -284,7 +274,7 @@ const GuidanceByStandby: React.FunctionComponent<{
                   ORDER BY greatest(age(backend_xmin), age(backend_xid)) DESC;`}
         />
       </CodeBlock>
-    </>
+    </li>
   );
 };
 
@@ -298,7 +288,7 @@ const GuidanceByPreparedXact: React.FunctionComponent<{
 
   const CodeBlock = useCodeBlock();
   return (
-    <>
+    <li>
       <h5>Abandoned prepared transactions</h5>
       <p>
         A transaction prepared for a two-phase commit will prevent cleanup until
@@ -331,7 +321,7 @@ const GuidanceByPreparedXact: React.FunctionComponent<{
                 ROLLBACK PREPARED <gid_from_above>;`}
         />
       </CodeBlock>
-    </>
+    </li>
   );
 };
 
--- a/components/CheckDocumentation/vacuum/XminHorizon.tsx
+++ b/components/CheckDocumentation/vacuum/XminHorizon.tsx
@@ -90,7 +90,12 @@ const GuidanceByBackend: React.FunctionComponent<{
     return null;
   }
 
-  const xminCode = <code>{xmin}</code>;
+  const xminBlock = xmin && (
+    <>
+      A long running transaction is holding back the xmin horizon at{" "}
+      <code>{xmin}</code>.
+    </>
+  );
   const CodeBlock = useCodeBlock();
   return (
     <>
@@ -101,8 +106,7 @@ const GuidanceByBackend: React.FunctionComponent<{
       </p>
       <h6>Solution</h6>
       <p>
-        {xmin &&
-          `A long running transaction is holding back the xmin horizon at ${xminCode}.`}
+        {xminBlock}
         You can find the transaction holding back the xmin horizon and its
         connection's pid by running the following command:
       </p>
--- a/components/CheckDocumentation/vacuum/XminHorizon.tsx
+++ b/components/CheckDocumentation/vacuum/XminHorizon.tsx
@@ -47,6 +47,7 @@ const XminHorizonGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
     byStandby = heldBackBy.find((v) => v["type"] === "standby");
     byPreparedXact = heldBackBy.find((v) => v["type"] === "prepared_xact");
   }
+  const inApp = heldBackBy === null;
   return (
     <div>
       <h4>Impact</h4>
@@ -61,33 +62,27 @@ const XminHorizonGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
         the unnecessary table bloat or slow queries.
       </p>
       <h4>Common Causes and Solutions</h4>
-      <GuidanceByBackend
-        inApp={!heldBackBy}
-        xmin={byBackend && byBackend["xmin"]}
-      />
+      <GuidanceByBackend inApp={inApp} xmin={byBackend && byBackend["xmin"]} />
       <GuidanceByReplicationSlot
-        inApp={!heldBackBy}
+        inApp={inApp}
         xmin={byReplicationSlot && byReplicationSlot["xmin"]}
         serverReplicationUrl={serverReplicationUrl}
       />
       <GuidanceByReplicationSlotCatalog
-        inApp={!heldBackBy}
+        inApp={inApp}
         xmin={byReplicationSlotCatalog && byReplicationSlotCatalog["xmin"]}
         serverReplicationUrl={serverReplicationUrl}
       />
-      <GuidanceByStandby
-        inApp={!heldBackBy}
-        xmin={byStandby && byStandby["xmin"]}
-      />
+      <GuidanceByStandby inApp={inApp} xmin={byStandby && byStandby["xmin"]} />
       <GuidanceByPreparedXact
-        inApp={!heldBackBy}
+        inApp={inApp}
         xmin={byPreparedXact && byPreparedXact["xmin"]}
       />
     </div>
   );
 };
 
-const GuidanceByPreparedXact: React.FunctionComponent<{
+const GuidanceByBackend: React.FunctionComponent<{
   inApp: boolean;
   xmin: number;
 }> = ({ inApp, xmin }) => {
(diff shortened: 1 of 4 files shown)


components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+      <ul>
+        <li>
+          <h5>Long-running transactions</h5>
+          <p>
+            When there is a long running transaction, you can't clean up rows
+            that the transaction can see, even if it already became a dead row.
+          </p>
+        </li>
+        <li>
+          <h5>Delayed/bad replication slots</h5>
```
msakrejda, line 73:
```suggestion
          <h5>Lagging or abandoned replication slots</h5>
```
Mostly a personal preference thing, but I don't like slashes in headings like this. And I think "lagging" is slightly more accurate than "delayed"  here, and "bad" is pretty vague. "Stale" could be another alternative to "abandoned".

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+          <p>
+            When the replication is delayed or the standby server is down, the
```
msakrejda, line 75:
```suggestion
            When replication is delayed or a replica server is down, the
```
No article needed here. And a replica is not necessarily a standby in this context, so we should use a more general term.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+            oldest transaction that the replication slot needs the database to
+            retain can be "stuck" hence hold back of the xmin horizon can
+            happen. This can also happen to the system catalogs.
```
msakrejda, line 78:
```suggestion
            retain can be "stuck", holding back the xmin horizon.
            This can also happen to the system catalogs.
```
Why are system catalogs special here? Should we explain this?

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+          </p>
+        </li>
+        <li>
+          <h5>Long-running transactions on standbys</h5>
+          <p>
+            When <code>hot_standby_feedback</code> is on, queries on standbys
```
msakrejda, line 84: We could use a SettingLink here, but that's probably not what we want because we'd be linking to the value of the setting on the primary, not on the replica, right? Just sanity-checking my own reasoning.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+            will act the same as primary regarding the xmin horizon, and causes
+            the xmin horizon to be behind.
```
msakrejda, line 86:
```suggestion
            will hold back the xmin horizon just as if they were running on the primary.
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+          </p>
+        </li>
+        <li>
+          <h5>Unfinished prepared transactions</h5>
```
msakrejda, line 90: Technically you may want to leave a prepared transaction around for a while and come back a few days later to commit it, but I agree that this is an unlikely use case, and we should treat it as such. Maybe "abandoned" is stronger than "unfinished" here?
```suggestion
          <h5>Abandoned prepared transactions</h5>
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+          <p>
+            If the transaction prepared for a two-phase commit is unfinished and
+            kept around, it can become the oldest xmin.
```
msakrejda, line 93:
```suggestion
            A transaction prepared for a two-phase commit will prevent cleanup until
            it is either committed or rolled back.
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+          </p>
+        </li>
+      </ul>
+      {heldBackBy && (
+        <>
+          <h4>Solution</h4>
+          {byBackend && (
+            <>
+              <p>
+                The long running transaction is holding back the xmin horizon at{" "}
```
msakrejda, line 103:
```suggestion
                A long running transaction is holding back the xmin horizon at{" "}
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+                <code>{byBackend["xmin"]}</code>. You can find such transaction
```
msakrejda, line 104:
```suggestion
                <code>{byBackend["xmin"]}</code>. You can find this transaction
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+                and its pid by running the following command:
```
msakrejda, line 105:
```suggestion
                and its connection's pid by running the following command:
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+                  inline
+                  sql={`SELECT pg_terminate_backend('<query_pid>');`}
+                />
+                .
+              </p>
+            </>
+          )}
+          {byReplicationSlot && (
+            <p>
+              The replication slot is holding back the xmin horizon at{" "}
```
msakrejda, line 129:
```suggestion
              A replication slot is holding back the xmin horizon at{" "}
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+              <code>{byReplicationSlot["xmin"]}</code>. This indicates that some
+              physical (streaming) replication slot is likely either delayed or
+              down. If the replication slot is no longer used, remove it with{" "}
```
msakrejda, line 132:
```suggestion
              physical (streaming) replication slot is likely either lagging or
              down. If the replication slot is no longer used, remove it with{" "}
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+              <SQL
+                inline
+                sql={`SELECT pg_drop_replication_slot('<slot_name>');`}
+              />
+              .
+            </p>
+          )}
+          {byReplicationSlotCatalog && (
+            <p>
+              The replication slot is holding back the xmin horizon at{" "}
```
msakrejda, line 142:
```suggestion
              A replication slot is holding back the xmin horizon at{" "}
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+              <code>{byReplicationSlotCatalog["xmin"]}</code>, specifically with
+              the system catalogs. This indicates that the replication slot is
```
msakrejda, line 144:
```suggestion
              system catalogs. This indicates that the replication slot is
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+              either delayed, down, or having trouble replicating due to schema
```
msakrejda, line 145:
```suggestion
              either lagging, down, or having trouble replicating due to schema
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+                inline
+                sql={`SELECT pg_drop_replication_slot('<slot_name>');`}
+              />
+              .
+            </p>
+          )}
+          {byStandby && (
+            <>
+              <p>
+                The long running transaction on the standby is holding back the
```
msakrejda, line 158:
```suggestion
                A long running transaction on a standby is holding back the
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+                  FROM pg_stat_replication
+                  ORDER BY age(backend_xmin) DESC;`}
+                />
+              </CodeBlock>
+            </>
+          )}
+          {byPreparedXact && (
+            <>
+              <p>
+                The orphaned prepared transaction is holding back the xmin
```
msakrejda, line 175:
```suggestion
                A prepared transaction is holding back the xmin
```
I think above it makes sense to talk about abandoned prepared transactions ("abandoned" probably makes more sense than "orphaned" since they don't necessarily have a parent in any meaningful sense), but here I don't think we want to imply that we have _identified_ this prepared transaction as something that definitely needs to be rolled back. It's possible that users actually want to commit it; we can let them make that call. I think the rest of the guidance can still focus on the likely "roll back" scenario.

msakrejda (commented):
A bunch more nitpicking, but looking great.

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: +1. I like stale more so I'll go with stale 👍

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: Uhmm since we are changing the title to "Lagging or stale replication slots", should we update here too, to something like:

> When replication is lagging or a replica server is stale (e.g. down, gone)

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: Yeah you're right, linking to primary's `hot_standby_feedback` here is meaningless. However, I think you have a good point re: linking to some settings, or we actually should be able to eliminate some suggestions based on how the current server is like.

I'll look into more for this one, I've actually been confused around here and it would be great to get the complete understanding, as well as showing better/proper explanation here.

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: this suggestion reads better with a new title, thanks 👍

keiko713:
(this one is not ready for re-review yet, still have things that I want to improve/address feedback)

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,205 @@
...
+      <p>VACUUM is potentially blocked by the xmin horizon.</p>
+      <p>
+        The xmin horizon tells you until which point the vacuum process can
+        clean up dead rows. When this value is behind and not making progress,
+        it's very likely that VACUUM can't clean up dead rows, hence blocking
+        VACUUM.
+      </p>
+      <p>
+        When VACUUM is blocked and dead rows can't be cleaned, it can result to
+        the unnecessary table bloat or slow queries.
```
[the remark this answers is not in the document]
keiko713: Uhmm, do you mean that copying that slow queries section here?

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: (I did that)

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: I've made a big change around here so it should be easier to understand. marking resolved for this specific one

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: We actually should be able to eliminate all unrelated suggestions, so I changed the structure.

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: FYI I change this to "the" actually, as I wanted to show this sentence within the docs site too (without mentioning the one sentence before).

keiko713:
@msakrejda during today's 1:1, Lukas and I had conversation around here and we decided to make a pretty big change to this, so now it's looking quite different. The main point is the same, but it would be great to get a review with fresh eyes.
I can also walk through this during tomorrow's sync meeting.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+import { useSmartAnchor } from "../../SmartAnchor";
+
+const XminHorizonTrigger: React.FunctionComponent<CheckTriggerProps> = ({
+  config,
+}) => {
+  const threshold = config.settings["behind_hours"] as number;
+  const hourPluralized = threshold === 1 ? "hour" : "hours";
+  return (
+    <p>
+      Detects when the xmin horizon on the server was assigned at more than{" "}
```
msakrejda, line 19:
```suggestion
      Detects when the xmin horizon on the server was assigned more than{" "}
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+      <code>{threshold}</code> {hourPluralized} ago and creates an issue with
+      severity "info". The issue will be created even if no VACUUM is currently
+      blocked by this, as this will potentially block any future VACUUMs.
+      Resolves once the xmin horizon is no longer behind.
```
msakrejda, line 23: Does "no longer behind" mean "under the threshold" here?

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+    );
+    byStandby = heldBackBy.find((v: any) => v["type"] === "standby");
+    byPreparedXact = heldBackBy.find((v: any) => v["type"] === "prepared_xact");
+  }
+  const inApp = issue != null;
+  const causeTitle = inApp
+    ? heldBackBy.length > 1
+      ? "Identified Causes and Solutions"
+      : "Identified Cause and Solution"
+    : "Common Causes and Solutions";
```
msakrejda, line 55: Nice.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+    <div>
+      <h4>Impact</h4>
+      <p>VACUUM is potentially blocked by the xmin horizon.</p>
+      <p>
+        The xmin horizon tells you up to which point the vacuum process can
+        clean up dead rows. When this value is behind and not advancing, VACUUMs
+        will be blocked and will not be able to clean up dead rows.
+      </p>
+      <p>
+        When VACUUM is blocked and dead rows can't be cleaned, it can result to
```
msakrejda, line 66:
```suggestion
        When VACUUM is blocked and dead rows can't be cleaned, it can result in
```

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+                  FROM pg_stat_activity
+                  WHERE backend_xmin IS NOT NULL OR backend_xid IS NOT NULL
+                  ORDER BY greatest(age(backend_xmin), age(backend_xid)) DESC;`}
+        />
+      </CodeBlock>
+      <p>You can cancel it by running either of commands:</p>
+      <CodeBlock>
+        <SQL
+          sql={`SELECT pg_cancel_backend('<query_pid>');
+                SELECT pg_terminate_backend('<query_pid>');`}
```
msakrejda, line 130: I feel like giving users a choice here but not explaining the choice is not a good pattern to get into. An explanation of the difference between these two would be kind of a tangent, and I think `pg_cancel_backend` works pretty reliably in modern Postgres, so I think we should only mention that. (It would also be okay to only mention `pg_terminate_backend`; I do think mentioning both is potentially confusing.)

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+  const Link = useSmartAnchor();
+  const CodeBlock = useCodeBlock();
+  return (
+    <li>
+      <h5>Lagging or stale physical replication slots</h5>
+      <p>
+        With physical streaming replication with{" "}
+        <code>hot_standby_feedback</code> is on, when replication is lagging or
+        a replica server is stale (e.g. down, gone), the oldest transaction that
```
msakrejda, line 154:
```suggestion
        a replica server is stale (e.g. down), the oldest transaction that
```
I think "gone" is pretty vague and not that different from "down". One example is probably enough to give users the general idea here.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+  const Link = useSmartAnchor();
+  return (
+    <li>
+      <h5>Lagging or stale logical replication slots</h5>
+      <p>
+        With logical replication slots, replication can also get stale when DDL
+        changes (database migrations) don't get applied to a replica server
+        (subscriber). When the subscriber was unable to replicate data due to a
+        schema mismatch, replication will error and get stale. This causes the
+        system catalogs xmin of the primary (publisher) to be held back until
```
msakrejda, line 201:
```suggestion
        xmin of the system catalogs of the primary (publisher) to be held back until
```
I think the triple compound noun here is kind of awkward. I admit the rewording is not great either, but I think it's a small step up. If you can think of other ways to reword that, that can work, too.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -0,0 +1,342 @@
...
+    return null;
+  }
+
+  const CodeBlock = useCodeBlock();
+  return (
+    <li>
+      <h5>Abandoned prepared transactions</h5>
+      <p>
+        A transaction prepared for a two-phase commit will prevent cleanup until
+        it is either committed or rolled back.
```
msakrejda, line 300: Since prepared transactions are not very common, and it's possible the DBA who is alerted about this issue is not familiar with them (even though the application programmer who caused the problem must have been), should we link to, e.g., https://www.postgresql.org/docs/current/sql-prepare-transaction.html somewhere here?

msakrejda (approved):
Some minor comments, but looks great.

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: I was like "I'm pretty positive I got this feedback before" and realized that it was with my blog post 🙃 I'll add a link 👍

components/CheckDocumentation/vacuum/XminHorizon.tsx
msakrejda: Hmm, yeah, I suppose. Maybe "once the age of the xmin horizon is under the threshold"? I think using "once the xmin horizon is under the threshold" as a shorthand also sounds okay to me if you think the above is too verbose.

keiko713: I'll go with "once the age of the xmin horizon is under the threshold" 👍

keiko713: 🤦 just realized that I forgot to update the "is under the threshold" part. I'm going to make a follow-up PR

keiko713: https://github.com/pganalyze/pganalyze-docs/pull/162
Decided to make even longer, making it "assignment age". This matches well with the graph title.

PR #155, opened 2023-05-30
Add VACUUM: Bloat - Insufficient VACUUM Frequency check

keiko713:
UnnecessaryTableGrowth uses a similar pattern as MissingIndex.
The docs doesn't have the solution section as it's in dynamic guidance generated in app side.

image


components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <p>
+      Detects when a table has experienced unnecessary growth due to
+      insufficient VACUUM. It analyzes the table statistic from the past seven
+      days and calculates the amount of growth that could have been avoided if
+      VACUUM had been performed more frequently. Creates an issue when the
+      unnecessary growth exceeds{" "}
+      <code>{formatBytes(Number(config.settings["notify_bytes"]))}</code>, as
+      well as <code>{config.settings["notify_pct"]}%</code> of the current size.
+      Resolves once the unnecessary growth decreases due to changes in
+      autovacuum settings or table usage.
```
msakrejda, line 23: Technically it resolves no matter what the reason for the decreased growth, no? Should we changes this to just
> Resolves once the unnecessary growth decreases.

? We can add more details in Guidance if necessary.

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <div>
+      <h4>Impact</h4>
+      <p>
+        In Postgres, any DELETEs or UPDATEs create dead rows that can be
+        VACUUMed to become reusable space for future INSERTs or UPDATEs.
+        Autovacuum is designed to automatically perform VACUUM regularly to
+        efficiently reclaim reusable space created by removing dead rows. This
+        helps ensure that INSERTs and UPDATEs can use existing space instead of
+        claiming new space. In this way, the table can avoid unnecessary growth.
```
msakrejda, line 42: I think we should emphasize the problem more strongly before talking about the mitigation. How about something like this?

```suggestion
        In Postgres, any DELETEs or UPDATEs create dead rows in the tables they modify.
        VACUUM allows Postgres to reclaim these rows and reuse that space for future INSERTs or UPDATEs.
        Autovacuum is designed to perform VACUUM regularly in order to
        efficiently reuse space. This
        helps ensure that INSERTs and UPDATEs can use existing space instead of
        claiming new space. In this way, the table can avoid unnecessary growth.
```

I also reworded a bit for conciseness and to avoid reusing "reclaim" (which I think is more useful in the second sentence).

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </p>
+      <p>
+        However, if autovacuum is not performed with the appropriate frequency
```
msakrejda, line 45:
```suggestion
        However, if VACUUM is not performed with the appropriate frequency
```

I think VACUUM is the important thing here, right? It doesn't matter whether it's through autovacuum or any other means.

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
```diff
@@ -0,0 +1,80 @@
...
+        or at the right time, dead rows may not be reclaimed. This can cause
+        unnecessary table growth, which could have been avoided if dead rows
+        were reclaimed. This typically result in table bloat where the physical
+        size of the table on disk is larger than its actual data size. Table
+        bloat is problematic for several reasons:
```
msakrejda, line 50:
```suggestion
        or at the right time, dead rows may not be reclaimed. This can cause
        unnecessary table growth and results in table bloat, where the physical
        size of the table on disk is larger than its actual data size. Table
        bloat is problematic for several reasons:
```

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </p>
+      <ul>
+        <li>
+          <h5>Slower queries</h5>
+          <p>
+            When a table is bloated, it takes more time to read and write data
+            from the disk, which can slow down queries that access the table.
```
msakrejda, line 57: I like the breakdown between query impact and system impact, but I think we could be more explicit here. E.g., something like
```suggestion
            Since bloated tables take more physical space for the same amount of actual data,
            they take longer to read from disk, and take up more room when cached in memory.
            This makes queries on these tables less efficient, because they have to do more
            work to scan the same amount of data.
```

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </p>
+        </li>
+        <li>
+          <h5>Increased disk usage</h5>
+          <p>
+            Table bloat can cause the disk space usage to increase
+            unnecessarily, which can lead to disk space shortages and degraded
+            performance.
```
msakrejda, line 65: I think the last part is a little vague. And I don't think we need to equivocate about the impact: except in some corner cases, bloat will always have a negative impact (no?). Maybe something like
```suggestion
            Table bloat causes disk space usage to increase
            unnecessarily. It also leads to more I/O to read and write the
            same amount of actual data.
```

msakrejda (commented):
Nice! Some nitpicky comments on the docs, but it's looking good.

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
keiko713: 👍 I'll tweak "reclaim these rows" to "reclaim these dead rows" (I was able to read smoother with it as I didn't need to think about what "these rows" were)

components/CheckDocumentation/vacuum/UnnecessaryTableGrowth.tsx
keiko713: Uhmm I kinda like the first sentence of your suggestion of "slower queries"

> Since bloated tables take more physical space for the same amount of actual data

I wonder if we can switch the order of "Increased disk usage" and "Slower queries", so that above sentence's "since" would read better. I'll try that version, let me know what you think.
Only downside is, I think query impact is more important than the system impact, so having the system impact item before the query impact might not be the ideal.

keiko713: Actually I'll keep the order, then change the title from "Increased disk usage" to "Increased disk usage and I/O" to show that it's more system oriented

keiko713:
Applied reviews, ready for another 👀

util/checks/index.ts
```diff
@@ -223,7 +226,16 @@ export const DEFAULT_CHECK_CONFIGS = {
...
-      threshold_count: 1,
+      settings: {
+        threshold_count: 1,
+      },
+    },
+    unnecessary_table_growth: {
+      enabled: true,
+      settings: {
+        notify_pct: 20,
+        notify_bytes: 10485760,
```
msakrejda, line 237: This is just ten megabytes? Is that right?

msakrejda (approved):

util/checks/index.ts
keiko713: Yup! I suppose I could do 10 * 1024 * 1024 actually, easier to read

msakrejda: Sorry, I meant that it seems like a fairly aggressive threshold (i.e., I think it might generate issues for a lot of tables with only modest bloat). Did you try this on our staging data? I don't think it's totally unreasonable, it just jumped out at me.

seanlinsley: 10 GB sounds reasonable to me. That would give us just a handful of alerts.

Screenshot 2023-06-09 at 11 34 39 AM

msakrejda: For what it's worth, the check (at least as documented--I have not reviewed the code yet) fires when _both_ the size threshold and percentage threshold are met. But 10MB still seems aggressive to me, even considering that.

msakrejda: So, thinking about this some more, a lot of bloat on a small table can be a problem, and there are use cases for single-row (or other very small, single-page) tables. Maybe this threshold is okay (as long as it's _both_ 20% and 10MB). I think it's probably still a good idea to sanity-check this against our own data.

keiko713: I just want to clarify what I'm setting as a threshold (as I just found out that my code was looking at the wrong value 🤦 gonna fix later) -- this threshold is the percentage & bytes of the unnecessary table growth, the growth that could have avoided with more vacuums. This means, it's actually totally unrelated to the current bloat size or % of bloat (we have a discussion to add such alert separately, but not for this one).
In other words, even if the current bloat is big, it's possible that it won't create this issue if that bloat was created long time ago and the table is not creating a bloat actively currently (because in that case, adjusting autovacuum setting won't help the situation).

With that, here is the list of current number with our data. Previously, I've seen that some partitioned tables were creating issues (not this time though, as all are under threshold).

* query_stats_35d_20230606, 7.0 %, 5327.38 MB
* schema_table_scan_methods, 10.0 %, 4019.46 MB (likely this and above are having big "could have avoided bytes" due to high inserts activity, which could make this value less accurate)
* schema_index_stats_hourlies_35d_20230609, 1.0 %, 43.73 MB
* query_stats_hourlies_35d_20230607, 7.0 %, 216.91 MB
* schema_table_stats_hourlies_35d_20230609, 1.0 %, 34.85 MB
* query_tagset_keys, 39.0 %, 8.33 MB
* users, 15.0 %, 0.95 MB
* api_keys, 11.0 %, 0.49 MB

In our case, I think threshold is working pretty well. For larger tables, the percentage correctly (?) helps to avoid triggering. For high percentage tables (`query_tagset_keys`), the size is also helping well. I can see that the size can be increased here though, to something like 100MB?

keiko713:
Gonna actually merge this as the app side will be merged too. We can always come back and update if we want to.

PR #156, opened 2023-06-06
2023.06.0 Enterprise release changelog, and related changes

lfittl:
See commit messages


enterprise/releases/2023-06-0.mdx
```diff
@@ -0,0 +1,77 @@
...
+- Allow setting `SIDEKIQ_CONCURRENCY=0` to turn off workers in combined Enterprise image
+  - You can utilize this setting in order to separate the workers from the container servicing the pganalyze web UI, avoiding slow web requests caused by worker processing
+  - Note such a setup requires having configured object storage as well as an external Redis server (set via `REDIS_URL`)
+  - See [all settings](/docs/enterprise/settings)
+
+
+## Performance
+
+- Query analysis: Speed up analysis by avoiding to process schema dependency tracking
+- Improve performance of getIssues by avoiding N+1 queries
```
keiko713, line 43:
```suggestion
- Improve performance of `getIssues` API endpoint by avoiding N+1 queries
```

I wonder if we can make this `getIssues` more clear that it's an API endpoint, something along that line. At minimum maybe adding backquote?

enterprise/releases/2023-06-0.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+- Object storage: Fix DNS lookup errors from Rust-based snapshot workers
+  - This may have shown as an I/O error that caused no snapshot data to be processed, despite the collector successfully sending data to the installation
+  - If we have previously asked you to utilize an IP address for your MinIO endpoint as a workaround, you can now switch back to using a DNS hostname
+- Query analysis:
+  - Don't retry query analysis after 'relation does not exist' errors
+  - Avoid "Unexpected rtekind" errors for functions and VALUES
+  - Fix planning for hash indexes (this allows Index Advisor to run on tables that have hash indexes, but does not yet recommend hash indexes)
+- Remove query length limit for parsing
+  - Previously query texts exceeding 20,000 characters were not parsed during the initial intake, causing them to not have referenced tables, or an associated statement type. This is change removes this limit.
```
keiko713, line 64:
```suggestion
  - Previously query texts exceeding 20,000 characters were not parsed during the initial intake, causing them to not have referenced tables, or an associated statement type. This is change removes this limit
```

Looks like this one only ends with period, let's remove it for consistency

enterprise/releases/2023-06-0.mdx
keiko713: Ah, well I just saw "getIssues API" section in bug fixes. I guess it's okay not to make this as backquote (I also realized that we don't use it much in changelog anyways).
Maybe instead, we say "Improve performance of getIssue API by avoiding N+1 queries"?

enterprise/releases/2023-06-0.mdx
```diff
@@ -0,0 +1,77 @@
...
+- getIssues API: Fix legacy displayName field to actually return the name, not the UUID
+  - Note that the displayName field is deprecated and replaced by `references`.`name` (see the current [getIssues API](/docs/api/queries/getIssues) docs)
+- Explain text to JSON: Add support for "Presorted Key" property
```
keiko713, line 67: Can you explain which PR is associated with this? I think I'm not so aware of this change and I don't really know which one you're talking about (though it might just be lack of product knowledge of me)

enterprise/releases/2023-06-0.mdx
```diff
@@ -0,0 +1,77 @@
...
+  - Batch check-up logic together for 10 database servers at a time
+
+
+## Bugfixes
+
+- Object storage: Fix DNS lookup errors from Rust-based snapshot workers
+  - This may have shown as an I/O error that caused no snapshot data to be processed, despite the collector successfully sending data to the installation
+  - If we have previously asked you to utilize an IP address for your MinIO endpoint as a workaround, you can now switch back to using a DNS hostname
+- Query analysis:
+  - Don't retry query analysis after 'relation does not exist' errors
```
keiko713, line 60: I _think_ we tend to use double quotes for things like this

```suggestion
  - Don't retry query analysis after "relation does not exist" errors
```

keiko713 (approved):
I have small notes but overall looking good!

enterprise/releases/2023-06-0.mdx
msakrejda: I think it's https://github.com/pganalyze/pganalyze/pull/2783 -- but "Explain text to JSON" may not mean much to our users. Maybe instead something like

> Fix parsing of explain plans using the "Presorted Key" property when using `auto_explain.log_format = text`"

?

keiko713: Ohh, thanks for the pointer. Yeah I think Maciek's proposal is more easy to understand, so +1 in general.

`auto_explain.log_format` is `text` by default so maybe it's okay to drop "when using..." part? Or, "when using text format auto_explain"?
Also technically it's a fix indeed (as it's introduced 13 and it's been a while), I think "add support" is not so wrong either. So something like:

> Add support for parsing of explain plans using the "Presorted Key" property
> Fix parsing of explain plans using the "Presorted Key" property when using text format auto_explain

(just throwing more suggestions, I'm fine with Maciek's suggestion too)

enterprise/releases/2023-06-0.mdx
lfittl: Good idea on spelling it out - I'll go with " `getIssues` API endpoint" in both places where its mentioned.

enterprise/releases/2023-06-0.mdx
lfittl: Yep, good catch - will remove that last sentence and the trailing dot.

enterprise/releases/2023-06-0.mdx
lfittl: Went with Maciek's suggestion - thanks!

msakrejda: > auto_explain.log_format is text by default so maybe it's okay to drop "when using..." part?

That's the default format, but we do recommend setting it to JSON since the text format is pretty ad-hoc and our transformation is not perfect. I'm -0.1 on "when using text format auto_explain": it flows better, but I don't think that's critical in a changelog, and I do think it's nice to mention the concrete setting we're referring to. And I do think "fix" makes more sense than "add support" here, since it was not something we knew didn't work.

(I know this is all moot now since this was just merged; just trying to hone our collective changelog game :smile: ).

PR #157, opened 2023-06-07
Fix structure of Inefficient Index Phase default check config

msakrejda:
Also add type declarations to the default check config structure to
avoid similar problems in the future, and fix a type definition error
with a helper function.


keiko713 (approved):
nice improvement 👍

PR #158, opened 2023-06-16
Add docs for pg_stat_statements_reset with collector hang

keiko713:
We've seen numbers of customers having this issue, so it's good to document this somewhere.

You can review in https://pganalyze-we-update-doc-vyocdc.herokuapp.com/docs/install/troubleshooting/collector_hang too


install/troubleshooting/collector_hang.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+```sql
+SELECT
+  count(*),
+  current_setting('pg_stat_statements.max'),
+  pg_size_pretty(sum(length(query)))
+FROM pg_stat_statements;
+```
+
+To mitigate the issue, you can run `SELECT pg_stat_statements_reset()` to reset the statistic data.
```
msakrejda, line 25:
```suggestion
To mitigate the issue, you can run `SELECT pg_stat_statements_reset()` to reset the statistics data.
```

I think it would also be good to mention something like "This will not affect the statistics data tracked within pganalyze, but if you are using other tools that rely on `pg_stat_statements` data, you should review how they will be affected."

install/troubleshooting/collector_hang.mdx
```diff
@@ -0,0 +1,48 @@
...
+When the collector test run (`pganalyze-collector --test --verbose`) is already hanging, it's good to try this out and see the test will stop hanging.
```
msakrejda, line 26:
```suggestion
If the collector test run (`pganalyze-collector --test --verbose`) is already hanging, it's good to try this out and see if it resolves the problem.
```

install/troubleshooting/collector_hang.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+If the collector still hangs after this, please [contact us](/contact) so that we can help debugging further.
```
msakrejda, line 28:
```suggestion
If the collector still hangs after this, please [contact us](/contact) so that we can help debug further.
```

install/troubleshooting/collector_hang.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+Once you confirm that `pg_stat_statements_reset()` is a valid mitigation, you can enable a recurring reset executed by pganalyze.
+
+## Setting up recurring `pg_stat_statements_reset()`
+
+First, you need to create the following helper method:
```
msakrejda, line 34:
```suggestion
First, identify the primary monitored database. This is the value of the `db_name` collector setting, or the first database in the list, if there are several. Then, you need to create the following helper method:
```

install/troubleshooting/collector_hang.mdx
```diff
@@ -0,0 +1,48 @@
...
+  /* pganalyze-collector */ SELECT * FROM public.pg_stat_statements_reset();
+$$ LANGUAGE sql VOLATILE SECURITY DEFINER;
+```
+
+Second, go to the Server Settings panel in Settings page and tune "Collector: Query Statistics Reset Frequency":
+
+![Server Settings](/server_settings.png)
+
+Please feel free to [reach out to us](/contact) for which frequency to pick.
+You could always start from 24 hours to see if that helps the situation.
```
msakrejda, line 48: I think if people do reach out, we'll probably just tell them to start at 24h and make it more frequent if necessary, so might as well start with that here:
```suggestion
Start with a reset frequency of 24 hours to see if that helps the situation. If this helps for a while but the problem eventually happens again, try increasing the frequency. Please feel free to [reach out to us](/contact) to tune it further.
```

msakrejda (approved):
A few minor comments, but looks great.

install/troubleshooting/collector_hang.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+```sql
+SELECT
+  count(*),
+  current_setting('pg_stat_statements.max'),
+  pg_size_pretty(sum(length(query)))
+FROM pg_stat_statements;
+```
+
+To mitigate the issue, you can run `SELECT pg_stat_statements_reset()` to reset the statistics data.
```
keiko713, line 25:
```suggestion
To mitigate the issue, you can run `SELECT pg_stat_statements_reset()` to reset the statistics data.
This will not affect the statistics data tracked within pganalyze, but if you are using other tools that rely on `pg_stat_statements` data, you should review how they will be affected.
```

PR #160, opened 2023-07-05
Add VACUUM Advisor doc base and Bloat page

keiko713:
This PR adds a super basic base of VACUUM Advisor doc, and add a new page "VACUUM Advisor: Bloat in Postgres".
To avoid showing up this in the menu, don't add contents in the main ToC or page.

https://pganalyze-we-vacuum-blo-9skjjl.herokuapp.com/docs/vacuum-advisor/bloat

### DIFF
--- a/vacuum-advisor/bloat-in-postgres.mdx
+++ b/vacuum-advisor/bloat-in-postgres.mdx
@@ -1,0 +1,79 @@
+---
+title: 'Bloat in Postgres'
+backlink_href: /docs/vacuum-advisor
+backlink_title: 'VACUUM Advisor'
+---
+
+In Postgres, bloat refers to a situation when the physical size of a table on
+disk is larger than its actual data size.
+
+## Why does bloat happen?
+
+In Postgres, any DELETEs or UPDATEs create dead rows in the tables they modify.
+This is due to [the MVCC mechanism](https://www.postgresql.org/docs/current/mvcc-intro.html)
+that Postgres uses to support concurrent transactions.
+VACUUM allows Postgres to reclaim these dead rows and reuse that space for
+future INSERTs or UPDATEs. Autovacuum is designed to perform VACUUM regularly in
+order to efficiently reuse space. This helps prevent unnecessary table growth.
+
+### Insufficient VACUUMs
+
+If VACUUM is not performed with the appropriate frequency or at the right time,
+dead rows may not be reclaimed, or may not be reclaimed quickly enough. This can
+cause the table to grow, using new disk space even though there is available
+unused space, since that unused space is still occupied by dead rows.
+
+### Inefficient VACUUMs
+
+Even when VACUUM is performed sufficiently, it's possible that dead rows still
+cannot be reclaimed. This can happen when transactions are still accessing those
+dead rows, and the xmin horizon — representing the point up to which VACUUM can
+clean up dead rows — is held back.
+As a result, VACUUM may not reclaim enough dead rows, leading to bloat.
+
+
+## Why is bloat a problem?
+
+Table bloat has several negative implications for the performance of your database.
+
+### Slower queries
+
+Since bloated tables take more physical space for the same amount of actual data,
+they take longer to read from disk, and take up more room when cached in memory.
+This makes queries on these tables less efficient, because they have to do more
+work to scan the same amount of data.
+
+### Increased disk space and I/O
+
+Table bloat unnecessarily consumes additional disk space and requires more I/O
+operations to read and write the same amount of actual data.
+
+
+## How to avoid bloat
+
+It is normal to have some bloat for tables with DELETEs or UPDATEs operations.
+However, you want to pay attention when you see an increasing trend in bloat.
+It is important to tune the VACUUM configuration to ensure VACUUMs are keeping
+up with bloat.
+It is also important to monitor your database and take action when the xmin
+horizon is held back, to avoid inefficient VACUUMs.
+pganalyze automatically performs checks to identify such cases, enabling you to
+easily take appropriate actions.
+
+* [Insufficient VACUUM Frequency](/docs/checks/vacuum/insufficient_vacuum_frequency)
+* [VACUUM Blocked By Xmin Horizon](/docs/checks/vacuum/xmin_horizon)
+
+
+## How to clean up existing bloat
+
+Once bloat has accumulated, adjusting the frequency or efficiency of VACUUM
+alone will not eliminate it. Bloat can only be eliminated when the space
+occupied by dead rows becomes reusable through VACUUM, and there are more
+INSERTs than DELETEs to utilize that space.
+
+To clean up existing bloat, you have two options: `VACUUM FULL` or
+[pg_repack](https://pganalyze.com/blog/5mins-postgres-pg-repack-VACUUM-FULL).
+However, `VACUUM FULL` requires heavyweight locks and must rewrite the entire
+table, which makes it impractical for most workloads. On the other hand,
+pg_repack rewrites the table without taking long locks and it is recommended
+approach over `VACUUM FULL`.
--- a/directory.json
+++ b/directory.json
@@ -875,6 +875,10 @@
     "vacuum-advisor/performance": {
       "title": "pganalyze VACUUM Advisor: Performance",
       "path": "/docs/vacuum-advisor/performance"
+    },
+    "vacuum-advisor/bloat-in-postgres": {
+      "title": "Bloat in Postgres",
+      "path": "/docs/vacuum-advisor/bloat-in-postgres"
     }
   },
   "logClassifications": {
--- a/vacuum-advisor/toc.yml
+++ b/vacuum-advisor/toc.yml
@@ -12,3 +12,5 @@
     href: vacuum-advisor/performance
   - name: "Activity"
     href: vacuum-advisor/activity
+  - name: "Bloat in Postgres"
+    href: vacuum-advisor/bloat-in-postgres


vacuum-advisor/index.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: 'VACUUM Advisor'
+backlink_href: /docs
+backlink_title: 'Documentation'
+---
+
+The pganalyze VACUUM Advisor is a tool to help tuning VACUUM.
```
msakrejda, line 7:
```suggestion
The pganalyze VACUUM Advisor is a tool to help tune VACUUM.
```
or
```suggestion
The pganalyze VACUUM Advisor is a tool to help tune VACUUM settings.
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
+---
+title: 'VACUUM Advisor: Bloat in Postgres'
+backlink_href: /docs/vacuum-advisor
+backlink_title: 'VACUUM Advisor'
+---
+
+In Postgres, bloat refers to a situation that the physical size of a table on
```
msakrejda, line 7:
```suggestion
In Postgres, bloat refers to a situation when the physical size of a table on
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+disk is larger than its actual data size.
+
+## Why does bloat happen?
+
+In Postgres, any DELETEs or UPDATEs create dead rows in the tables they modify.
```
msakrejda, line 12: I think we should mention MVCC somehow. We don't need to explain it at length, but if you have no idea what that is or why INSERTs or UPDATEs would create dead rows, we can at least give you something to look up. Something like
```suggestion
In Postgres, any DELETEs or UPDATEs create dead rows in the tables they modify. This is due to the MVCC mechanism that Postgres uses to support concurrent transactions.
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+VACUUM allows Postgres to reclaim these dead rows and reuse that space for
+future INSERTs or UPDATEs. Autovacuum is designed to perform VACUUM regularly in
+order to efficiently reuse space. This helps ensure that INSERTs and UPDATEs can
+use existing space instead of claiming new space. This helps prevent unnecessary
+growth of tables by utilizing existing space.
```
msakrejda, line 17:
```suggestion
order to efficiently reuse space. This helps prevent unnecessary
table growth.
```
I think this could be more compact. Those extra words aren't saying anything.

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+### Insufficient VACUUMs
+
+If VACUUM is not performed with the appropriate frequency or at the right time,
+dead rows may not be reclaimed, causing the table to grow using new disk space
+even though there is available unused space.
```
msakrejda, line 23: In contrast, I think this could be more explicit. I think it would also flow better split up into separate sentences:

```suggestion
If VACUUM is not performed with the appropriate frequency or at the right time,
dead rows may not be reclaimed, or may not be reclaimed quickly enough. This can cause
the table to grow, using new disk space even though there is available unused space, since
that unused space is still occupied by dead rows.
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+### Inefficient VACUUMs
+
+Even when VACUUM is performed sufficiently, it's possible that dead rows still
+cannot be reclaimed. This can happen when transactions are still accessing those
+dead rows, and the xmin horizon — represents the point up to which point VACUUM
+can clean up dead rows — is held back.
```
msakrejda, line 30:
```suggestion
dead rows, and the xmin horizon — representing the point up to which VACUUM
can clean up dead rows — is held back.
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+As a result, VACUUM may not reclaim any or only a few dead rows, leading to bloat.
```
msakrejda, line 31:
```suggestion
As a result, VACUUM may not reclaim enough dead rows, leading to bloat.
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+
+## Why bloat is so bad
```
msakrejda, line 34: Technically this works as a statement rather than a question, but I think having parallel construction in the titles (starting with "Why...") but having one as a question and another as a statement is kind of awkward. Also I think we should use a phrase a little more formal than "is so bad" :smile: . Maybe something like

```suggestion
## Why is bloat a problem?
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+Table bloat has several negative implications, primarily affecting the
+performance of your database.
```
msakrejda, line 37: It doesn't affect anything else, does it? If you have huge bloat problems, you probably have freezing problems, too, but we should cover that separately.
```suggestion
Table bloat has several negative implications for the performance of your database.
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+### Increased disk space and I/O
+
+Table bloat unnecessarily consumes additional disk space and requires more I/O
+operations to read and write the same amount of actual data.
+
+
+## How to avoid bloat
+
+It is important to tune the VACUUM configuration to avoid insufficient VACUUMs.
```
msakrejda, line 54: I think the double negative here is kind of akward. Maybe
```suggestion
It is important to tune the VACUUM configuration to ensure VACUUMs are managing bloat.
```
or
```suggestion
It is important to tune the VACUUM configuration to ensure VACUUMs are keeping up with bloat.
```
?

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+It is also important to monitor your database and taking action when the xmin
```
msakrejda, line 55:
```suggestion
It is also important to monitor your database and take action when the xmin
```

vacuum-advisor/bloat.mdx
```diff
@@ -0,0 +1,76 @@
...
+horizon is held back, to avoid Inefficient VACUUMs.
```
msakrejda, line 56:
```suggestion
horizon is held back, to avoid inefficient VACUUMs.
```

msakrejda (approved):
Some wording comments, but looks great. I wonder if we should also talk about "it's okay to have some bloat". We show bloat numbers in a bunch of places and we consciously decided not to flag them as good or bad. I think that was the right idea, as the estimates are derived using some gross approximations, but it may leave the users without a solid understanding of how to reason about these numbers.

vacuum-advisor/bloat.mdx
keiko713: Thanks for this suggestion, I didn't like that title and but I wasn't able to come up something nice, so I was really seeking some suggestion 🙈

vacuum-advisor/bloat.mdx
keiko713: I think we could say that increased disk space it also about the capacity problem. Though it's a tiny detail so I'm fine simplifying 👍

vacuum-advisor/index.mdx
keiko713: Note: this section will be overwritten by Lukas, it's more as a placeholder. I'll update though

keiko713:
New link: https://pganalyze-we-vacuum-blo-9skjjl.herokuapp.com/docs/vacuum-advisor/bloat-in-postgres

> I wonder if we should also talk about "it's okay to have some bloat". We show bloat numbers in a bunch of places and we consciously decided not to flag them as good or bad. I think that was the right idea, as the estimates are derived using some gross approximations, but it may leave the users without a solid understanding of how to reason about these numbers.

Good point. We could add something in some section (either section would be a good fit). What do you think about something like the following?

> It is normal to have some bloat for tables with DELETEs or UPDATEs operations. However, you want to pay attention when you see an increase trend of bloat.

We could add something like this in the beginning of "How to avoid bloat" section, for example.

msakrejda:
>> It is normal to have some bloat for tables with DELETEs or UPDATEs operations. However, you want to pay attention when you see an increase trend of bloat.

That looks good, though maybe "...when you see an increasing trend in bloat."

vacuum-advisor/bloat.mdx
msakrejda: Ah good point. Yeah, I think we can probably gloss over disk space.

vacuum-advisor/bloat-in-postgres.mdx
```diff
@@ -0,0 +1,79 @@
...
+However, you want to pay attention when you see an increasing trend in bloat.
+It is important to tune the VACUUM configuration to ensure VACUUMs are keeping
+up with bloat.
+It is also important to monitor your database and take action when the xmin
+horizon is held back, to avoid inefficient VACUUMs.
+pganalyze automatically performs checks to identify such cases, enabling you to
+easily take appropriate actions.
+
+* [Insufficient VACUUM Frequency](/docs/checks/vacuum/insufficient_vacuum_frequency)
+* [VACUUM Blocked By Xmin Horizon](/docs/checks/vacuum/xmin_horizon)
```
keiko713, line 64: maybe we can actually link to the vacuum-advisor/bloat page from here
welp I'll just merge this as is though

PR #163, opened 2023-07-12
Add VACUUM: Freezing - Approaching Transaction/Multixact ID Wraparound check doc

keiko713:
Two checks are really similar, but slightly different.

### DIFF
--- a/components/CheckDocumentation/vacuum/MxidWraparound.tsx
+++ b/components/CheckDocumentation/vacuum/MxidWraparound.tsx
@@ -30,25 +30,44 @@ const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
     <div>
       <h4>Impact</h4>
       <p>
-        The multixact ID is close to wraparound. This is about a multixact ID,
-        not about the normal transaction ID.
+        Multixact ID space utilization is high and approaching to wraparound.
       </p>
       <p>
-        A multixact ID is an internal identifier used support row locking by
-        multiple transactions. Multixact IDs are created when transactions use{" "}
-        <code>SELECT ... FOR UPDATE</code> to lock and update tuples.
+        This is about a multixact ID, not about the normal transaction ID. A
+        multixact ID is an internal identifier used support row locking by
+        multiple transactions. Multixact IDs are commonly used with{" "}
+        <code>SELECT ... FOR UPDATE</code> query.
       </p>
       <p>
-        When the multixact ID wraparound happens, Postgres is unable to start a
-        new multixact and essentially makes the database down. Postgres usually
-        runs anti-wraparound autovacuums to prevent this from happening, but if
-        such VACUUMs are either not running or are running but unable to freeze
-        rows for some reasons, the transaction age (and the utilization) keeps
-        growing and eventually hit the point of wraparound. With Postgres 14+,
-        there is a special type of failsafe VACUUM which takes extraordinary
-        measures to avoid system-wide multixact ID wraparound failure. It is
-        recommended to take any actions to prevent the wraparound, or even
-        failsafe from happening.
+        Postgres runs autovacuums regularly and this helps keeping multixact ID
+        space utilization low by freezing old multixact IDs. There is also the
+        anti-wraparound autovacuum specific for freezing, when the regular
+        autovacuums are either not freezing or simply not running. This
+        anti-wraparound autovacuum will be triggered when utilization (age)
+        exceeds the threshold, specified in{" "}
+        <code>autovacuum_multixact_freeze_max_age</code>.
+      </p>
+      <p>
+        In order to avoid wraparound failure, old multixact IDs must be freezed
+        by VACUUMs and this is not avoidable. The more old multixact IDs need to
+        be freezed, the more expensive VACUUM costs, potentially causes the
+        overall performance degradation or simply takes really long time to
+        finish. The anti-wraparound autovacuum holds{" "}
+        <code>SHARE UPDATE EXCLUSIVE</code> lock, which can block DDL
+        statements.
+      </p>
+      <p>
+        If multixact ID space utilization reaches to 99.85% (3M multixact left),
+        the system will shut down and refuse to start any new multixact. This is
+        to prevent any data corruption from happening by running out multixact
+        ID. Resolving this requires manual intervention.
+      </p>
+      <p>
+        With Postgres 14+, there is a special type of failsafe VACUUM which
+        takes extraordinary measures to avoid the shutdown. While this is very
+        useful, you still want to avoid this as a failsafe autovacuum ignores
+        resource utilization constraints and can have significant performance
+        impact.
       </p>
       <h4>Common Causes</h4>
       <ul>
@@ -58,18 +77,14 @@ const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
             Autovacuums, especially anti-wraparound autovacuums, are meant to
             freeze old multixact IDs and keep utilization low in order to
             prevent wraparound from happening. When autovacuum is turned off, or
-            the setup of anti-wraparound autovacuum is not fitting well for the
-            database usage, it is possible that autovacuum is not working as
-            intended and causing the utilization growth. When autovacuum is
-            turned off, or autovacuum settings are not well suited to actual
-            database usage, it is possible that autovacuum is not able to keep
-            up with freezing old multixact IDs and causes utilization to grow.
-            Make sure that autovacuum is turned on, and revisit the
-            configuration settings to ensure that autovacuum will keep multixact
-            ID space utilization under control. You can check out the
-            configuration settings related to freezing in{" "}
-            <Link to={serverVacuumFreezingUrl}>Freezing</Link> page in VACUUM
-            Advisor.
+            autovacuum settings are not well suited to actual database usage, it
+            is possible that autovacuum is not able to keep up with freezing old
+            multixact IDs and causes utilization to grow. Make sure that
+            autovacuum is turned on, and revisit the configuration settings to
+            ensure that autovacuum will keep multixact ID space utilization
+            under control. You can check out the configuration settings related
+            to freezing in <Link to={serverVacuumFreezingUrl}>Freezing</Link>{" "}
+            page in VACUUM Advisor.
           </p>
         </li>
         <li>
@@ -81,8 +96,9 @@ const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
             typical blocker is a long running transaction, or a transaction
             holding some locks that could cancel autovacuums. It is important to
             make sure that there is no such transactions so that VACUUMs can
-            make freezing progress. Check for any long-running transactions on
-            the <Link to={backendsUrl}>Connections</Link> page.
+            make freezing progress. Check for any long-running transactions or
+            its lock state on the <Link to={backendsUrl}>Connections</Link>{" "}
+            page.
           </p>
         </li>
       </ul>
@@ -92,7 +108,7 @@ const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
 
 const documentation: CheckDocs = {
   description:
-    "Alerts when the multixact ID space utilization on the server exceeds the specified percentage.",
+    "Alerts when multixact ID space utilization on the server exceeds the specified percentage.",
   Trigger: MxidWraparoundTrigger,
   Guidance: MxidWraparoundGuidance,
 };
--- a/components/CheckDocumentation/vacuum/MxidWraparound.tsx
+++ b/components/CheckDocumentation/vacuum/MxidWraparound.tsx
@@ -1,0 +1,92 @@
+import React from "react";
+
+import {
+  CheckDocs,
+  CheckGuidanceProps,
+  CheckTriggerProps,
+} from "../../../util/checks";
+import { useSmartAnchor } from "../../SmartAnchor";
+
+const MxidWraparoundTrigger: React.FunctionComponent<CheckTriggerProps> = ({
+  config,
+}) => {
+  const warning = config.settings["warning_threshold_pct"] as number;
+  const critical = config.settings["critical_threshold_pct"] as number;
+  return (
+    <p>
+      Detects when the multixact ID utilization on the server is more than{" "}
+      <code>{warning}%</code> and creates an issue with severity "warning".
+      Escalates to "critical" once the utilization exceeds{" "}
+      <code>{critical}%</code>. Resolves once the utilization falls below the
+      threshold.
+    </p>
+  );
+};
+
+const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
+  urls: { backendsUrl },
+  issue,
+}) => {
+  const Link = useSmartAnchor();
+  return (
+    <div>
+      <h4>Impact</h4>
+      <p>
+        The multixact ID is close to wraparound. This is about a multixact ID,
+        not about the normal transaction ID.
+      </p>
+      <p>
+        A multixact ID is an internal identifier used support row locking by
+        multiple transactions. Multixact IDs are created when transactions use{" "}
+        <code>SELECT ... FOR UPDATE</code> to lock and update tuples.
+      </p>
+      <p>
+        When the multixact ID wraparound happens, Postgres is unable to start a
+        new multixact and essentially makes the database down. Postgres usually
+        runs anti-wraparound autovacuums to prevent this from happening, but if
+        such VACUUMs are either not running or are running but unable to freeze
+        rows for some reasons, the transaction age (and the utilization) keeps
+        growing and eventually hit the point of wraparound. With Postgres 14+,
+        there is a special type of failsafe VACUUM which takes extraordinary
+        measures to avoid system-wide multixact ID wraparound failure. It is
+        recommended to take any actions to prevent the wraparound, or even
+        failsafe from happening.
+      </p>
+      <h4>Common Cases</h4>
+      <li>
+        <h5>Ineffectual autovacuum settings</h5>
+        <p>
+          Autovacuums, especially anti-wraparound autovacuums are meant to
+          prevent the wraparound from happening. When autovacuum is turned off,
+          or the setup of anti-wraparound autovacuum is not fitting well for the
+          database usage, it is possible that autovacuum is not working as
+          intended and causing the utilization growth. Make sure that autovacuum
+          is turned on, and revisit the setting to ensure that anti-wraparound
+          autovacuums will run periodically.
+        </p>
+      </li>
+      <li>
+        <h5>Blocked autovacuums</h5>
+        <p>
+          Even though autovacuum settings are adequate, it is possible that
+          autovacuums themselves are blocked by something hence freezing (to
+          reduce the multixact ID utilization) is not happening. The typical
+          blocker is a long running transaction, or a transaction holding some
+          locks that could cancel autovacuums. It is important to make sure that
+          there is no such transactions so that VACUUMs can freeze rows well.
+          Check for any long-running transactions on the{" "}
+          <Link to={backendsUrl}>Connections</Link> page.
+        </p>
+      </li>
+    </div>
+  );
+};
+
+const documentation: CheckDocs = {
+  description:
+    "Alerts when the multixact ID utilization on the server exceeds the specified percentage.",
+  Trigger: MxidWraparoundTrigger,
+  Guidance: MxidWraparoundGuidance,
+};
+
+export default documentation;
--- a/components/CheckDocumentation/vacuum/MxidWraparound.tsx
+++ b/components/CheckDocumentation/vacuum/MxidWraparound.tsx
@@ -25,7 +25,6 @@ const MxidWraparoundTrigger: React.FunctionComponent<CheckTriggerProps> = ({
 
 const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
   urls: { backendsUrl },
-  issue,
 }) => {
   const Link = useSmartAnchor();
   return (
--- a/components/CheckDocumentation/vacuum/MxidWraparound.tsx
+++ b/components/CheckDocumentation/vacuum/MxidWraparound.tsx
@@ -14,17 +14,16 @@ const MxidWraparoundTrigger: React.FunctionComponent<CheckTriggerProps> = ({
   const critical = config.settings["critical_threshold_pct"] as number;
   return (
     <p>
-      Detects when the multixact ID utilization on the server is more than{" "}
+      Detects when multixact ID space utilization on the server is more than{" "}
       <code>{warning}%</code> and creates an issue with severity "warning".
-      Escalates to "critical" once the utilization exceeds{" "}
-      <code>{critical}%</code>. Resolves once the utilization falls below the
-      threshold.
+      Escalates to "critical" once utilization exceeds <code>{critical}%</code>.
+      Resolves once utilization falls below the threshold.
     </p>
   );
 };
 
 const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
-  urls: { backendsUrl },
+  urls: { backendsUrl, serverVacuumFreezingUrl },
 }) => {
   const Link = useSmartAnchor();
   return (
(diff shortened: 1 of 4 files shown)


components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { useSmartAnchor } from "../../SmartAnchor";
+
+const MxidWraparoundTrigger: React.FunctionComponent<CheckTriggerProps> = ({
+  config,
+}) => {
+  const warning = config.settings["warning_threshold_pct"] as number;
+  const critical = config.settings["critical_threshold_pct"] as number;
+  return (
+    <p>
+      Detects when the multixact ID utilization on the server is more than{" "}
```
msakrejda, line 17:
```suggestion
      Detects when multixact ID utilization on the server is more than{" "}
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <code>{warning}%</code> and creates an issue with severity "warning".
+      Escalates to "critical" once the utilization exceeds{" "}
```
msakrejda, line 19:
```suggestion
      Escalates to "critical" once utilization exceeds{" "}
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <code>{critical}%</code>. Resolves once the utilization falls below the
```
msakrejda, line 20:
```suggestion
      <code>{critical}%</code>. Resolves once utilization falls below the
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+const MxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
+  urls: { backendsUrl },
+}) => {
+  const Link = useSmartAnchor();
+  return (
+    <div>
+      <h4>Impact</h4>
+      <p>
+        The multixact ID is close to wraparound. This is about a multixact ID,
+        not about the normal transaction ID.
```
msakrejda, line 35:
```suggestion
        The multixact ID counter is close to wraparound. This concerns multixact IDs,
        not normal transaction IDs.
```

I think there are two distinct concepts: TXIDs themselves, and the counter assigning TXIDs (or MXIDs in this case). It's a subtle difference, but I think it's helpful to distinguish between these here.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </p>
+      <p>
+        A multixact ID is an internal identifier used support row locking by
+        multiple transactions. Multixact IDs are created when transactions use{" "}
+        <code>SELECT ... FOR UPDATE</code> to lock and update tuples.
```
msakrejda, line 40: Is this the only case where MXIDs are used? I don't think we have to be exhaustive here, but if these could be coming from other places, we should avoid implying that this is the one source, since users may get confused if they have MXID problems but are _not_ using <code>SELECT ... FOR UPDATE</code>.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </p>
+      <p>
+        When the multixact ID wraparound happens, Postgres is unable to start a
+        new multixact and essentially makes the database down. Postgres usually
```
msakrejda, line 44: Wraparound doesn't really happen, though. Maybe  something like
```suggestion
        When multixact ID wraparound is very close, Postgres is unable to execute commands that require multixact IDs to be assigned
        and shuts down to avoid data corruption. Postgres usually
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+        runs anti-wraparound autovacuums to prevent this from happening, but if
+        such VACUUMs are either not running or are running but unable to freeze
+        rows for some reasons, the transaction age (and the utilization) keeps
+        growing and eventually hit the point of wraparound. With Postgres 14+,
+        there is a special type of failsafe VACUUM which takes extraordinary
+        measures to avoid system-wide multixact ID wraparound failure. It is
+        recommended to take any actions to prevent the wraparound, or even
+        failsafe from happening.
```
msakrejda, line 52: How about something like
```suggestion
        such VACUUMs are either not running, or are running but unable to freeze
        rows fast enough, the transaction age (and the utilization) keeps
        growing and eventually approaches the point of wraparound. With Postgres 14+,
        there is a special type of failsafe VACUUM which takes extraordinary
        measures to avoid system-wide multixact ID wraparound failure. Since
        a shutdown to prevent wraparound requires manual intervention, and
        a failsafe autovacuum ignores resource utilization constraints and can have
        significant performance impact, it is critical to ensure that freezing proceeds
        normally and mutlixact ID age is kept in check.
```
? I think this makes it a little more concrete.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <li>
+          <h5>Ineffectual autovacuum settings</h5>
+          <p>
+            Autovacuums, especially anti-wraparound autovacuums are meant to
+            prevent the wraparound from happening. When autovacuum is turned
+            off, or the setup of anti-wraparound autovacuum is not fitting well
+            for the database usage, it is possible that autovacuum is not
+            working as intended and causing the utilization growth. Make sure
+            that autovacuum is turned on, and revisit the setting to ensure that
+            anti-wraparound autovacuums will run periodically.
```
msakrejda, line 65: I think we should focus on how freezing is sort of a race against database utilization here. Maybe something like

```suggestion
            Autovacuums, especially anti-wraparound autovacuums, are meant to
            reclaim old multixact IDs in order to prevent wraparound from happening.
            When autovacuum is turned off, or autovacuum settings are not well suited
            to actual database usage, it is possible that autovacuum is not able to keep
            up with reclaiming old multixact IDs and causes utilization to grow. Make sure
            that autovacuum is turned on, and revisit the configuration settings to ensure
            that autovacuum will keep multixact ID utilization under control.
```

I also dropped a number of "anti-wraparound" because in theory, regular vacuum should be doing this job. We don't want to give the impression that anti-wraparound vacuums are _required_ to prevent wraparound and that regular vacuums don't help with that.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+          </p>
+        </li>
+        <li>
+          <h5>Blocked autovacuums</h5>
+          <p>
+            Even though autovacuum settings are adequate, it is possible that
```
msakrejda, line 71:
```suggestion
            Even if autovacuum settings are adequate, it is possible that
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+            autovacuums themselves are blocked by something hence freezing (to
```
msakrejda, line 72:
```suggestion
            autovacuums themselves are blocked by something, hence freezing (to
```
Bonus points for "hence"! And "Ineffectual" above. :golf::clap:.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+            reduce the multixact ID utilization) is not happening. The typical
```
msakrejda, line 73:
```suggestion
            reduce multixact ID utilization) is not able to keep up. The typical
```
It could be happening, just not fast enough. (I think this distinction is important since users may see log messages about freezing progress and still run into problems.)

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,93 @@
...
+            blocker is a long running transaction, or a transaction holding some
+            locks that could cancel autovacuums. It is important to make sure
+            that there is no such transactions so that VACUUMs can freeze rows
+            well. Check for any long-running transactions on the{" "}
```
msakrejda, line 77:
```suggestion
            that there is no such transactions so that VACUUMs can make freezing progress.
            Check for any long-running transactions on the{" "}
```

msakrejda (commented):
I took a first pass on the mxid docs. I imagine the txid one is similar, and I think most of my comments apply there, too. Looks good, but I do think we should fine-tune the language a bit to avoid confusing users.

keiko713:
> I took a first pass on the mxid docs. I imagine the txid one is similar, and I think most of my comments apply there, too.

Oh noes, I should have mentioned here too, but they are pretty different, especially "impact" section as I did some rewrite with your comment in the other PR. I wanted to see which version you'll like more 🙈 
No rush but could you please take a look TXID one too? The trigger component or common causes section are still pretty similar so I can apply the current reviews 👍

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <div>
+      <h4>Impact</h4>
+      <p>
+        The transaction ID usage is high. Postgres usually runs anti-wraparound
+        autovacuums to keep the usage low by freezing old rows. However, if such
+        VACUUMs are either not running or are running but unable to freeze rows
+        for some reasons, the transaction usage keeps growing.
+      </p>
+      <p>
+        The anti-wraparound VACUUM needs to happen at some point and this is not
```
msakrejda, line 40: I think this confuses the issue: it's technically true that VACUUM needs to perform freezing (i.e., the "anti-wraparound" part), but this does not have to be the "vacuum (to prevent wraparound)" that doesn't cede locks. Ideally, all the freezing you need happens in the course of regular vacuums.

(Confusing the issue even further is that the current Postgres vacuum scheduling mechanism is poor, so in practice you'll often get into situations where you won't get freezing until an anti-wraparound vacuum happens, but that's a slightly different issue, and probably not something we need to get into here.)

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,97 @@
...
+        VACUUM becomes more and more expensive, potentially causes the overall
+        performance degradation or simply takes really long time to finish. The
+        anti-wraparound VACUUM holds <code>SHARE UPDATE EXCLUSIVE</code> lock,
+        which can block DDL statements.
+      </p>
+      <p>
+        When the transaction ID usage reaches to 99.85% (3M transactions left),
+        the system will shut down and refuse to start any new transactions. This
+        is to prevent any data corruption from happening by running out
+        transaction ID.
```
msakrejda, line 51: We don't need to get into the details of how to get out of this situation, but we should probably hint at the fact that this is not game over for your database. Maybe something like
```suggestion
        transaction ID. Resolving this requires manual intervention.
```

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,97 @@
...
+      </p>
+      <p>
+        With Postgres 14+, there is a special type of failsafe VACUUM which
+        takes extraordinary measures to avoid the shutdown. While this is very
```
msakrejda, line 55: We should talk about increased resource usage here (either specifically how it skips the cost limit system, or just in general warn about more aggressive resource usage.)

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,97 @@
...
+        useful, you still want to avoid even this from happening, to avoid any
+        impact by this VACUUM.
+      </p>
+      <h4>Common Cases</h4>
```
msakrejda, line 59:
```suggestion
      <h4>Common Causes</h4>
```

msakrejda (commented):
Okay, I saw your comment and took a very quick look at the txid section, too. This looks good, but I think we should still make some similar tweaks. I'll take a more thorough look tomorrow morning.

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,97 @@
...
+  );
+};
+
+const TxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
+  urls: { backendsUrl },
+}) => {
+  const Link = useSmartAnchor();
+  return (
+    <div>
+      <h4>Impact</h4>
```
msakrejda, line 32: Stepping back from commenting on specific copy, I think we want to discuss a few things here:
 - the TXID/MXID counter is approaching wraparound
 - vacuum normally addresses this through freezing but may not be keeping up
   - may want to mention anti-wraparound vacuums
 - if this is not addressed, the database will eventually shut down to prevent corruption and require manual intervention
 - Postgres 14+ gives you failsafe autovacuum, which is an additional safeguard against this

So I think the docs below are roughly the right structure, but:
 - "transaction ID usage" is sort of ambiguous: users could interpret that to mean allocation
   - we could use "transaction ID counter is approaching wraparound" as in my MXID suggestion above (this still uses the term "wraparound" which I had said I wanted to avoid in the description, but I think it's okay as long as we're not implying that wraparound will happen)
   - we could talk about the "transaction ID space utilization" -- this may be a little too abstract, but the Postgres docs do mention "xid space" and it's sometimes used to discuss these concepts
 - we should avoid implying that anti-wraparound vacuums are the only thing that help reclaim TXIDs/MXIDs

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <p>
+        The anti-wraparound VACUUM needs to happen at some point and this is not
+        avoidable. If VACUUM is somehow not run in timely manner, the upcoming
+        VACUUM becomes more and more expensive, potentially causes the overall
+        performance degradation or simply takes really long time to finish. The
+        anti-wraparound VACUUM holds <code>SHARE UPDATE EXCLUSIVE</code> lock,
+        which can block DDL statements.
+      </p>
+      <p>
+        When the transaction ID usage reaches to 99.85% (3M transactions left),
```
msakrejda, line 48:
```suggestion
        If the transaction ID usage reaches to 99.85% (3M transactions left),
```

Hopefully no one actually has to deal with a "when" here :slightly_smiling_face:.

msakrejda (commented):
Okay, added some more feedback. I'll be out for a bit during your early morning today but should be back by 4:30 Pacific. Let's try to sync on this and wrap it up if you have time.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
keiko713: per sync, I'll use "multixact ID space utilization" here too, I think it's better to use the same terminology here

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+}) => {
+  const Link = useSmartAnchor();
+  return (
+    <div>
+      <h4>Impact</h4>
+      <p>
+        Multixact ID space utilization is high and approaching to wraparound.
+      </p>
+      <p>
+        This is about a multixact ID, not about the normal transaction ID. A
```
msakrejda, line 36:
```suggestion
        This is about multixact IDs, not about normal transaction IDs. A
```
We're not talking about one specific one, so I think we want plurals here. Alternately, we could say
```suggestion
        This is about the multixact ID counter, not about normal the transaction ID counter. A
```
but I think that's more awkward. Or possibly 
```suggestion
        This is about the multixact ID space, not about normal the transaction ID space. A
```
I think any of these could work.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        multixact ID is an internal identifier used support row locking by
```
msakrejda, line 37:
```suggestion
        multixact ID is an internal identifier used to support row locking by
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        multiple transactions. Multixact IDs are commonly used with{" "}
+        <code>SELECT ... FOR UPDATE</code> query.
```
msakrejda, line 39:
```suggestion
        <code>SELECT ... FOR UPDATE</code> queries.
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+      </p>
+      <p>
+        Postgres runs autovacuums regularly and this helps keeping multixact ID
+        space utilization low by freezing old multixact IDs. There is also the
+        anti-wraparound autovacuum specific for freezing, when the regular
```
msakrejda, line 44:
```suggestion
        anti-wraparound autovacuum specifically for freezing, when the regular
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        autovacuums are either not freezing or simply not running. This
+        anti-wraparound autovacuum will be triggered when utilization (age)
+        exceeds the threshold, specified in{" "}
+        <code>autovacuum_multixact_freeze_max_age</code>.
+      </p>
+      <p>
+        In order to avoid wraparound failure, old multixact IDs must be freezed
```
msakrejda, line 51:
```suggestion
        In order to avoid wraparound failure, old multixact IDs must be frozen
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        by VACUUMs and this is not avoidable. The more old multixact IDs need to
+        be freezed, the more expensive VACUUM costs, potentially causes the
+        overall performance degradation or simply takes really long time to
+        finish. The anti-wraparound autovacuum holds{" "}
+        <code>SHARE UPDATE EXCLUSIVE</code> lock, which can block DDL
+        statements.
```
msakrejda, line 57: I think we can break this up into additional sentences to make it easier to follow. Also, I'd drop "this is not avoidable" (you said "must" above so this doesn't add anything and put the more common problem (really long runtime) first.

```suggestion
        by VACUUMs. The more old multixact IDs need to
        be frozen, the more expensive VACUUM becomes. It can take a really long
        time to finish and potentially cause overall performance degradation. The
        anti-wraparound autovacuum holds a <code>SHARE UPDATE EXCLUSIVE</code> lock,
        which can block DDL statements.
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        This is about a multixact ID, not about the normal transaction ID. A
+        multixact ID is an internal identifier used support row locking by
+        multiple transactions. Multixact IDs are commonly used with{" "}
+        <code>SELECT ... FOR UPDATE</code> query.
+      </p>
+      <p>
+        Postgres runs autovacuums regularly and this helps keeping multixact ID
+        space utilization low by freezing old multixact IDs. There is also the
+        anti-wraparound autovacuum specific for freezing, when the regular
+        autovacuums are either not freezing or simply not running. This
```
msakrejda, line 45:
```suggestion
        autovacuums are either not freezing aggressively enough or simply not running. This
```
I think "not freezing" may make users wonder "why would they not be freezing" whereas "not freezing aggressively enough" suggests a configuration issue.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        by VACUUMs and this is not avoidable. The more old multixact IDs need to
+        be freezed, the more expensive VACUUM costs, potentially causes the
+        overall performance degradation or simply takes really long time to
+        finish. The anti-wraparound autovacuum holds{" "}
+        <code>SHARE UPDATE EXCLUSIVE</code> lock, which can block DDL
+        statements.
+      </p>
+      <p>
+        If multixact ID space utilization reaches to 99.85% (3M multixact left),
+        the system will shut down and refuse to start any new multixact. This is
```
msakrejda, line 61:
```suggestion
        If multixact ID space utilization reaches to 99.85% (3M multixact IDs left),
        the system will shut down. This is
```
I think that second part is sort of implicit, and you clarify it in the following sentence anyway.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        to prevent any data corruption from happening by running out multixact
+        ID. Resolving this requires manual intervention.
```
msakrejda, line 63:
```suggestion
        IDs. Resolving this requires manual intervention.
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+        resource utilization constraints and can have significant performance
+        impact.
+      </p>
+      <h4>Common Causes</h4>
+      <ul>
+        <li>
+          <h5>Ineffectual autovacuum settings</h5>
+          <p>
+            Autovacuums, especially anti-wraparound autovacuums, are meant to
+            freeze old multixact IDs and keep utilization low in order to
```
msakrejda, line 78: I think we should still avoid "utilization" here. Maybe we can frame this the other way instead, like
```suggestion
            freeze old multixact IDs and keep sufficient multixact IDs available in order to
```

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+            prevent wraparound from happening. When autovacuum is turned off, or
+            autovacuum settings are not well suited to actual database usage, it
+            is possible that autovacuum is not able to keep up with freezing old
+            multixact IDs and causes utilization to grow. Make sure that
```
msakrejda, line 82: I also think we should try to avoid "utilization" here due to ambiguity.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+            autovacuum is turned on, and revisit the configuration settings to
+            ensure that autovacuum will keep multixact ID space utilization
```
msakrejda, line 84: This one is okay, since you're talking about the space rather than XIDs themselves.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <h5>Blocked autovacuums</h5>
+          <p>
+            Even if autovacuum settings are adequate, it is possible that
+            autovacuums themselves are blocked by something, hence freezing (to
+            reduce multixact ID space utilization) is not able to keep up. The
+            typical blocker is a long running transaction, or a transaction
+            holding some locks that could cancel autovacuums. It is important to
+            make sure that there is no such transactions so that VACUUMs can
+            make freezing progress. Check for any long-running transactions or
+            its lock state on the <Link to={backendsUrl}>Connections</Link>{" "}
```
msakrejda, line 100:
```suggestion
            make freezing progress. Check for any long-running transactions and
            their lock state on the <Link to={backendsUrl}>Connections</Link>{" "}
```

If there is a lock state, there's a long-running transaction, right? I guess you could have a transaction without a lock state, but since you can't have it the other way around, I think "and" is more natural unless we want to reword this more elaborately.

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+const TxidWraparoundGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
+  urls: { backendsUrl, serverVacuumFreezingUrl },
+}) => {
+  const Link = useSmartAnchor();
+  return (
+    <div>
+      <h4>Impact</h4>
+      <p>
+        Transaction ID space utilization is high and approaching to wraparound.
```
msakrejda, line 33:
```suggestion
        Transaction ID space utilization is high and approaching wraparound.
```

components/CheckDocumentation/vacuum/TxidWraparound.tsx
```diff
@@ -0,0 +1,110 @@
...
+        ID space utilization low by freezing old transaction IDs. There is also
+        the anti-wraparound autovacuum specific for freezing, when the regular
+        autovacuums are either not freezing or simply not running. This
+        anti-wraparound autovacuum will be triggered when utilization (age)
+        exceeds the threshold, specified in{" "}
+        <code>autovacuum_freeze_max_age</code>.
+      </p>
+      <p>
+        In order to avoid wraparound failure, old transaction IDs must be
+        freezed by VACUUMs and this is not avoidable. The more old transaction
```
msakrejda, line 46:
```suggestion
        frozen by VACUUMs. The more old transaction
```

msakrejda: Since the rest here is very similar to the above, I think the above comments apply.

msakrejda (approved):
Nice, I think this looks great now. I still have some scattered comments on details of the copy, but I think the overall structure works well.

components/CheckDocumentation/vacuum/MxidWraparound.tsx
keiko713: Fyi decided to drop "and causes utilization to grow". I think it can read okay without it

PR #168, opened 2023-07-21
Update TOC for checks doc

keiko713:
In addition to TOC update, I also updated the check name to use "of" and "by" in non-title case


lfittl (approved):
:+1: I think we can also start merging this and deploying this to production, since the launch is around the corner.

PR #174, opened 2023-07-28
API docs: Update addServer, add updateServer and deleteServer

lfittl:
We've recently added additional AWS options to addServer, so document them. Additionally document the updateServer and deleteServer actions, to allow better automation of which servers are registered in pganalyze.


keiko713 (approved):
nice, looking good 👍 I think we could improve the format of API documentation (e.g. required/optional/default will be next to the param name), but we can think about it in another time

PR #175, opened 2023-07-29
VACUUM Advisor: Group related documentation together by category in TOC

lfittl:
With the introduction of the "What is X?" pages did this become a bit hard to parse after all. To help, group together the VACUUM Advisor category page with the related explanation pages in the TOC.

In passing revise the page titles of the category pages to be worded "[category] in VACUUM Advisor" instead of
"pganalyze VACUUM Advisor: [category]", and use that wording for the TOC as well, if the category page is in a group. "Activity" is the exception here, as it remains a top-level TOC item under VACUUM Advisor.

Additionally, rename "Bloat in Postgres" to "What is Bloat?" to avoid the awkwardly similar name to "Bloat in VACUUM Advisor", and to match how we named "What is Freezing?".

---

Before:

Screenshot 2023-07-28 at 5 06 46 PM

After:

Screenshot 2023-07-28 at 5 06 36 PM


keiko713 (approved):
I think this grouping works 👍 

example link:
https://pganalyze-we-group-vacu-bgtith.herokuapp.com/docs/vacuum-advisor/bloat

Are we okay not changing the link of "What is Bloat?" from bloat-in-postgres to what-is-bloat?

lfittl:
> Are we okay not changing the link of "What is Bloat?" from bloat-in-postgres to what-is-bloat?

Yeah, we should go ahead and change it for consistency. Will do that.

PR #181, opened 2023-08-18
Add billing FAQ to explain a few common questions about billing

lfittl:
This also serves as a location we can link to to explain the upcoming "billable server" terminology.

### DIFF
--- a/accounts/billing.mdx
+++ b/accounts/billing.mdx
@@ -14,7 +14,7 @@ For example, a typical Amazon Aurora cluster will contain 1 writer instance, and
 
 ## How can we pay for pganalyze?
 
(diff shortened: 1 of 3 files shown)


msakrejda (approved):
Should we have an entry for discounts?

lfittl:
> Should we have an entry for discounts?

Good question - in general I feel like the more we talk about "discounts" the more people will ask for them (e.g. that's why the replica multiplier is worded as a multiplier, not a discount).

I feel like unless we opt to offer more discount programs (e.g. for early stage startups/OSS projects/etc), the one FAQ entry on volume discounts is a good starting point for now - since that's the main discount we offer.

PR #182, opened 2023-08-23
Create section for collector Data collection and Snapshot submission Frequency

keiko713:
Preview: https://pganalyze-we-collector--kdqcvc.herokuapp.com/docs/collector#data-collection-and-snapshot-frequency

I'll likely need to get eyes from someone who is familiar with the collector to make sure I'm writing the correct information.

@jensy Maybe you could review regarding if it's a good structure?

### DIFF
--- a/collector/index.mdx
+++ b/collector/index.mdx
@@ -31,7 +31,7 @@ as well as to receive database logs by serving as a Heroku log drain.
 
 ## Data collection and Snapshot submission Frequency
 
-The pganalyze collector gathers metrics and metadata through various snapshots, each with specific cadences.
+The pganalyze collector gathers metrics and metadata through various "snapshots", each with specific cadences.
 This section explains the types of data collected and when different snapshot types are sent.
 
 There are three different types of snapshots that are sent to the pganalyze app.
--- a/collector/index.mdx
+++ b/collector/index.mdx
@@ -29,12 +29,19 @@ The collector makes only outbound connections, it does not listen on any ports. 
 Heroku-based setup, where the collector listens on the dyno's PORT for Heroku to consider the application running,
 as well as to receive database logs by serving as a Heroku log drain.
 
-## Data collection and Snapshot Frequency
+## Data collection and Snapshot submission Frequency
 
 The pganalyze collector gathers metrics and metadata through various snapshots, each with specific cadences.
 This section explains the types of data collected and when different snapshot types are sent.
 
 There are three different types of snapshots that are sent to the pganalyze app.
+
+* **Compact Activity Snapshot**: Sent every 10 seconds
+* **Compact Logs Snapshot**:
+  * Streaming: Sent approximately every 10 seconds
+  * Download: Sent every 30 seconds
+* **Full Snapshot**: Sent every 10 minutes (logged as just "snapshot")
+
 Successful submissions are logged as follows:
 
 ```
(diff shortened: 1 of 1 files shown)


jensy:
Yay, I think this is fantastic! I wish we would adjust the article width based on browser size, so we could account better for the table, and have less wraps, but that is a different story.

LGTM!

collector/index.mdx
```diff
@@ -29,3 +29,33 @@ The collector makes only outbound connections, it does not listen on any ports.
...
+Submitted compact activity snapshot successfully
+Submitted compact logs snapshot successfully
+Submitted snapshot successfully
+```
+
+These snapshots include metrics and metadata collected according to specific schedules and serve the following purposes:
+
+| Data             | Collection Interval | Snapshot Type    | Features                    |
+|------------------|---------------------|------------------|-----------------------------|
+| Connections      | Every 10 seconds    | Compact Activity | Connections                 |
```
seanlinsley, line 57: To save horizontal space, I think we can drop "every" from the collection intervals.

keiko713: Yeah we could drop indeed. Jens and I talked about this a bit, here is how it looks like:

image

Given that dropping "Every" won't help much re: horizontal space, we are leaning towards to keep "Every" for every thing, to make this part clear.

collector/index.mdx
```diff
@@ -29,3 +29,33 @@ The collector makes only outbound connections, it does not listen on any ports.
...
+| Query Stats      | Every 1 minute      | Full             | Query Performance           |
+| Other Stats      | Every 10 minutes    | Full             | Schema Statistics, System, Index Advisor, VACUUM Advisor |
+| Logs (Streaming) | Every 10 seconds    | Compact Logs     | Log Insights, EXPLAIN plans |
```
keiko713, line 60: Just realized that I should also set Approx. or something for this one (e.g. "Approx. every 10 seconds")

keiko713:
Either @seanlinsley or @msakrejda (or both), could you please double check the accuracy of this article?

collector/index.mdx
```diff
@@ -29,3 +29,33 @@ The collector makes only outbound connections, it does not listen on any ports.
 Heroku-based setup, where the collector listens on the dyno's PORT for Heroku to consider the application running,
 as well as to receive database logs by serving as a Heroku log drain.
 
+## Data collection and Snapshot submission Frequency
+
+The pganalyze collector gathers metrics and metadata through various snapshots, each with specific cadences.
```
msakrejda, line 34:
```suggestion
The pganalyze collector gathers metrics and metadata through various "snapshots", each with specific cadences.
```
I think because we use the term "snapshot" in a very specific way when talking about the collector, we should put the first occurrence in quotes to indicate that this term has special meaning here.

collector/index.mdx
```diff
@@ -29,3 +29,33 @@ The collector makes only outbound connections, it does not listen on any ports.
...
+This section explains the types of data collected and when different snapshot types are sent.
+
+There are three different types of snapshots that are sent to the pganalyze app.
+
+* **Compact Activity Snapshot**: Sent every 10 seconds
+* **Compact Logs Snapshot**:
+  * Streaming: Sent approximately every 10 seconds
+  * Download: Sent every 30 seconds
```
msakrejda, line 42: If we're breaking this out, maybe we should explain when which kind of log collection happens...

collector/index.mdx
msakrejda: I think if we say "approximately" above, we can leave this as is to conserve some space.

msakrejda (approved):

msakrejda (commented):
Oh, and, all the technical content looks right to me :+1:.

collector/index.mdx
keiko713: Yeah, I agree on this, but honestly I haven't figured out the difference between them 🙃 I'm planning to do some follow-up around this area to explain better once I figured it out 👍

PR #183, opened 2023-09-13
Add a limitation docs for VACUUM Advisor xmin horizon graph

keiko713:
🍐ing PR: https://github.com/pganalyze/pganalyze/pull/3253


vacuum-advisor/index.mdx
```diff
@@ -33,3 +33,4 @@ The insights are enabled by default for all your database servers, but alerts ne
 * VACUUM Advisor does not distinguish between dead row versions and dead line pointers (LP_DEAD), e.g. as caused by HOT updates and on-access pruning
 * Insights for tuning `autovacuum_cost_delay`, `autovacuum_cost_limit` and `autovacuum_max_workers` are still under development
 * Email alerts are not yet enabled for VACUUM Advisor insights
+* In the "VACUUM Blocked by Xmin Horizon" graph, when the exact assigned time of the xmin horizon wasn't retrievable (e.g. it was assigned before the statistics retention period), it will use the oldest statistics collected time and can be inaccurate. For example, even if the xmin horizon was assigned 60 days ago, the xmin horizon in the graph may show that it was assigned 35 days ago
```
lfittl, line 36: This sounds like there are cases where the horizon wasn't retrievable that are not because it exceeds the statistics retention period.

Assuming the only case is because the horizon is too old, maybe we can reword like this to avoid confusion:

In the "VACUUM Blocked by Xmin Horizon" graph, when the assigned time of the xmin horizon is older than the the statistics retention period (35 days), it will use the oldest known statistics collected time. For example, even if the xmin horizon was assigned 60 days ago, the xmin horizon in the graph would show that it was assigned 35 days ago.

lfittl (approved):

vacuum-advisor/index.mdx
keiko713: Cool cool, I like the rewording, thanks for the suggestion 👍 
I'm slightly concerned to say that the retention period being 35 days over there (it's officially 30 days in pricing page, also 14 days in production), but I think it's okay to do so to simplify (also we mention 35 days here and there already).

PR #184, opened 2023-09-13
Update guided setup requirements in our docs

msakrejda:
In https://github.com/pganalyze/collector/pull/438 we updated host OS
requirements for the collector, but we did not update our guided setup
requirements in our instructions.

Update the docs to reflect the new requirements.


msakrejda:
Note also that our manual setup instructions do not list any host OS requirements. Probably a job for a different PR, but should we add some?

keiko713 (approved):

PR #186, opened 2023-09-14
Add documentation for getServers API endpoint

lfittl:


api/queries/getServers.mdx
```diff
@@ -0,0 +1,64 @@
...
+## Use case
+
+You can utilize `getServers` to get a list of all servers that are integrated with an organization in pganalyze, and to find servers that have not received any data recently due to collector integration issues.
+
+## API
+
+Arguments for `getServers`:
+
+* **`organizationSlug` (string)**<br/>Organization identifier ("slug") for which to retrieve servers for
+* **`withRecentData` (boolean)**<br/>Filters for servers that have received data in the last 24 hours (optional), defaults to no filtering - if set to `false` it will filter for all servers that exist, but don't have recent data
```
keiko713, line 16:
```suggestion
* **`withRecentData` (boolean)**<br/>Filters for servers that have received data in the last 24 hours (optional), defaults to no filtering - if set to `false` it will filter for servers that haven't received data in the last 24 hours
```

While it's exactly opposite of when `true` is set and might be obvious, I think this is a bit more clear.

keiko713 (approved):
Might be good to get another eyes but looks good 👍

msakrejda (approved):
I did not check for consistency with existing GraphQL endpoint docs, but otherwise looks good.

PR #187, opened 2023-09-15
Create docs for VACUUM simulator

keiko713:
Create docs for VACUUM/ANALYZE Activity page and VACUUM simulator.
In addition to the documentation itself, I also want folks to check if the category of docs, if under VACUUM Advisor makes sense. These pages are technically in Schema Statistics page, but I think grouping this under VACUUM Advisor instead of creating a new category "Schema Statistics".

I also would like to add some gif if possible for VACUUM simulator, but I was having hard time creating gif from mp4. If you all think it's worth adding it, let's think about how to add it. Here is mp4:

https://github.com/pganalyze/pganalyze-docs/assets/911433/95835098-8851-4488-974d-ffd3c15b62e2

Preview: https://pganalyze-we-update-doc-alw6bt.herokuapp.com/docs/vacuum-advisor/vacuum-simulator


vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+  the simulation, potentially cause the simulation to trigger less autovacuums
+  by freeze age.
+
+
+## Triggered by inserts
+
+Inserts autovacuum is introduced on Postgres 13.
+With an inserts only table, because it won't generate any dead rows, autovacuums
+won't be triggered by dead rows. It can still be triggered by freeze age, but
+depending on the database workload, it can be a long time until it's triggered.
```
msakrejda, line 141: I think we should lead with what insert-triggered autovacuum is (or why it exists), rather than what Postgres version introduced it. We should also point out why it's bad to wait a long time to freeze. And I think "append-only" is a more standard term than "inserts only"; we should probably stick to that.

```suggestion
Since an append-only table won't generate any dead rows, autovacuum
won't be triggered by dead rows. It can still be triggered by freeze age, but
depending on the database workload, it can be a long time until it's triggered.
This can turn normal incremental maintenance into a resource-intensive,
long-running task that can impact database performance. To counteract
this, since Postgres 13, autovacuum can be triggered by INSERTs.
```

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+Inserts autovacuum is triggered to freeze old transaction IDs like freeze age
```
msakrejda, line 142:
```suggestion
INSERT-based autovacuum is triggered to freeze old transaction IDs like freeze age
```
or
```suggestion
Autovacuum based on inserts is triggered to freeze old transaction IDs like freeze age
```

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+autovacuum, but more importantly, it is triggered to update a visibility map.
+Each table has a visibility map to keep track of which pages contain only tuples
+that are known to be visible to all active transactions. For the case like an
+inserts only table, all the pages of the table should be visible here and less
+likely to be not visible moving forward (will be flagged as not visible if there
+is any write operation on that page).
```
msakrejda, line 148:
```suggestion
that are known to be visible to all active transactions. For a case like an
append-only table, all pages of the table should be all-visible, and are
likely to stay all-visible moving forward (they will be flagged as not all-visible if there
is any write operation on that page).
```
I think it's important to talk about "all-visible" specifically, not just "visible", to keep things clear.

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+However, Postgres doesn't know about it until VACUUM runs to set a bit (flag).
+Updating the visibility map will help to spend less time for any subsequent
+VACUUMs as these pages don't need to be VACUUMed and will be skipped. Updating
```
msakrejda, line 151:
```suggestion
Updating the visibility map will improve performance for any subsequent
VACUUMs as all-visible pages don't need to be VACUUMed and will be skipped. Updating
```
We're kind of blurring the all-visible/all-frozen distinction here, but that's probably okay given the level of detail. I'm kind of torn: on the one hand, I think trying to add in that info might make this too complicated, but on the other it feels like this could be misleading otherwise. Thoughts?

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+the visibility map also helps to use index-only scans, as this information will
+be used to determinate weather to use index-only scans or not.
```
msakrejda, line 153: I think index-only scans can still be used without visibility map information, they just have to do extra table reads, right?

```suggestion
the visibility map also helps index-only scans, as they will be able to avoid reading
the table data to check visibility, improving performance.
```

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+threshold, inserts autovacuums will be triggered:
+
+```
+threshold = autovacuum_vacuum_insert_threshold + autovacuum_vacuum_insert_scale_factor * number of tuples (pg_class.reltuples, number of live rows)
+```
+
+With VACUUM Simulator, you can tweak `autovacuum_vacuum_insert_threshold` and
+`autovacuum_vacuum_insert_scale_factor` in this section.
+Lowering these configuration settings triggers autovacuums more with some
+inserts only or inserts heavy table, though this rarely triggers autovacuums for
```
msakrejda, line 167:
```suggestion
Lowering these configuration settings triggers autovacuums more with
append-only or append-mostly tables, though this rarely triggers autovacuums for
```

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+other types of tables.
+
+If you are observing too many inserts autovacuums with the table, you can make
+`autovacuum_vacuum_insert_scale_factor` 0 and adjust
+`autovacuum_vacuum_insert_threshold` to find a reasonable threshold for inserts
+autovacuums.
+
+### Limitation
```
msakrejda, line 175:
```suggestion
### Limitations
```

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+
+* This simulation is only available with servers running Postgres 13 or above.
+* With the simulation, inserts autovacuum freezes old transaction IDs as it can
+  do it.
+  However, in reality, it's less likely that inserts autovacuum will freeze old
+  transaction IDs.
```
msakrejda, line 181:
```suggestion
* With the simulation, INSERT-based autovacuum freezes old transaction IDs as it can.
  However, in reality, it's less likely that it will freeze old transaction IDs.
```

Why less likely? I think this is also something we should try to explain better.

vacuum-advisor/what-is-freezing.mdx
```diff
@@ -15,7 +15,7 @@ about the ID of the transaction that created it. This is due to [the MVCC
 mechanism](https://www.postgresql.org/docs/current/mvcc-intro.html) that
 Postgres uses to support concurrent transactions. Since Postgres uses a 32-bit
 circular transaction ID space, those older IDs must be reused eventually. In
-order to avoid amibguity about which concrete transaction an ID refers to (known
+order to avoid ambiguity about which concrete transaction an ID refers to (known
```
msakrejda, line 18: Nice catch.

msakrejda (approved):
Sorry for the delayed review--I lost track of this one. Some wording suggestions and some comments and questions, but looks great.

msakrejda:
Oh, and I don't think the gif is necessary: we're not doing that anywhere else, and I think the docs and screenshots give a good overview of the tool.

lfittl:
> Oh, and I don't think the gif is necessary: we're not doing that anywhere else, and I think the docs and screenshots give a good overview of the tool.

As an alternate idea, we could also make a short walkthrough video on YouTube and embed that - but we could consider that separately from this PR.

vacuum-advisor/table-activity.mdx
```diff
@@ -0,0 +1,66 @@
...
+on but set to a higher value, or Log Insights is not enabled.
+
+In this list, you can check out the progress of freezing old transaction IDs or
+cleaning up dead tuples (dead rows). Both processes should occur with each
+VACUUM run; however, it is possible that no progress occurs under the following
+situations:
+
+* Freezing progress: when VACUUM was focusing to clean up dead tuples
+* Freezing progress: there was nothing can be frozen (often true in autovacuums triggered by inserts)
+* Dead tuple progress: more dead tuple created while VACUUM was running
```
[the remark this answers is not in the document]
keiko713: Actually I changed this part with the new PR I opened. I'll merge with the following (so that I can rebase this to make it purely for vacuum simulator) but let me know how it looks, happy to tweak more:

> * No freezing progress: the current oldest unfrozen transaction ID age is under `vacuum_freeze_min_age`
> * No dead tuple progress: more dead tuples were created while VACUUM was running

----

> Can there also be no dead tuple progress when freezing must happen but there are no dead tuples?

I'm thinking, we can either leave this out (as this is more obvious case, that there is no dead tuples == nothing to clean), or we can have this + "there was nothing that could be frozen" together to be more explicit. I'm leaning toward leaving this out.

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+
+There is a specific insight for this:
+[Bloat - Insufficient VACUUM Frequency](/docs/checks/vacuum/insufficient_vacuum_frequency).
+
+### Limitation
+
+* The simulation data tends to trigger autovacuum earlier than the actual
+  triggered timing, especially with tables with lots of UPDATEs.
+  Occasionally, UPDATEs operations don't create dead rows, or even decrease dead
+  rows when "mini autovacuum" happened with these operations.
```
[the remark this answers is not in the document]
keiko713: I actually kinda forgot where I got this wording from, but I think I got it from our 5 mins postgres: https://pganalyze.com/blog/5mins-postgres-debug-UPDATE-bloated-tables-auto-explain-pageinspect#hot-pruning-in-postgres

> Postgres can actually do a “mini-VACUUM” of sorts, where it just vacuums the page that you're modifying, because it’s already modifying that page anyway.

I'll put the link to there

vacuum-advisor/vacuum-simulator.mdx
keiko713: Uhmmm good point. After re-reading code:

* freezing vacuum: can't skip all-visible pages (but may skip all-frozen pages)
* dead rows and inserts vacuum: can skip all-visible pages

So "all-visible pages don't need to be VACUUMed" is not 100% accurate. We could add "unless it's a freezing (aggressive) vacuum" (I've been avoiding to use "aggressive vacuum" but I think I should be using as it is commonly used) to make it more accurate without introducing the word of "all-frozen".

Alternatively, inserts vacuum likely flag the page as both all-visible and all-frozen, so we can say something like:

> Updating the visibility map helps index-only scans, as they will be able to
avoid reading the table data to check visibility, improving performance.
> During the VACUUM of an append-only table, it is often able to mark pages as all-frozen, in addition to all-visible, and it'll stay that way moving forward too. This will improve performance for any subsequent VACUUMs as all-visible and all-frozen pages don't need to be VACUUMed and will be skipped.

Thoughts?

vacuum-advisor/vacuum-simulator.mdx
keiko713: I'm actually removing this, I wasn't understanding this behavior well enough -- freezing was skipped because the current oldest unfrozen transaction ID age was under `vacuum_freeze_min_age`. I'm planning to make a logic change in the simulator so we should be able to remove this limitation.

vacuum-advisor/vacuum-simulator.mdx
```diff
@@ -0,0 +1,182 @@
...
+value and graph section to show how it is based on the collected statistics and
+the adjustable configuration settings.
+You can also create a permalink of your current simulation and modified settings
+to share with your team members.
+
+### Limitation
+
+* The current autovacuum count can be much bigger than the simulation count with
+  frequently autovacuumed tables, since the simulation can only perform
+  autovacuum every 10 minutes.
```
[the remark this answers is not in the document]
keiko713: Uhmm yeah, I understand your point, though I don't have good alternative.

I think people can still use simulation for these tables to:

* understand what is causing so many vacuums
* find out how they can reduce autovacuum count

How does mentioning these sound?

keiko713:
@msakrejda thank you so much for your thorough review! I think I've updated most of them but left some comments for the parts that I want more of your inputs.
This PR now only include the vacuum simulator docs, the other part is already merged.

vacuum-advisor/vacuum-simulator.mdx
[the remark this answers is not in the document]
msakrejda: Yeah, that sounds reasonable. Maybe something like
```
* The current autovacuum count can be much bigger than the simulation count with
  frequently autovacuumed tables, since the simulation can only perform
  autovacuum every 10 minutes. The simulation can still be useful to understand why
  so many vacuums are performed, and to explore ways to reduce autovacuum frequency.
```
?

vacuum-advisor/vacuum-simulator.mdx
msakrejda: Yeah, that makes sense. Just minor wording tweaks:

>Updating the visibility map helps index-only scans, as they will be able to
>avoid reading table data to check visibility, improving performance.
>During VACUUM of an append-only table, it is often able to mark pages as all-frozen, in addition to all-visible, and it will stay that way (since these rows will never be modified). This will improve performance for any subsequent VACUUMs as pages that are both all-visible and all-frozen don't need to be VACUUMed and will be skipped.

(The most important is rewording that last sentence to avoid ambiguity.)

vacuum-advisor/vacuum-simulator.mdx
msakrejda: Yeah, I think you're right: the live rows fraction is taken into account when costing IOSes. I don't think we need to talk about that, though, and we can focus on general IOS performance.

msakrejda (approved):

PR #188, opened 2023-09-29
Enterprsie S3 object storage: Clarify where the configuration is set

lfittl:
There was an occurrence where someone accidentally thought these are set in the collector configuration. Given our recent standardization around always using "/etc/pganalyze.env" as the source for Docker environment settings when running in a VM, reference that here explicitly in an effort to help understand the instructions better.

In passing link to Enterprise Server container settings page when talking about the settings for static AWS credentials.


seanlinsley (approved):
👍. btw, the commit message misspells "enterprise"

PR #189, opened 2023-09-29
Install overview: Refresh logos, add CosmosDB to Azure DB for Postgres

lfittl:
Before:

Screenshot 2023-09-29 at 10 29 48 AM

After:

Screenshot 2023-09-29 at 10 49 13 AM


style.module.scss
```diff
@@ -90,6 +90,7 @@
     justify-content: space-between;
     display: flex;
     align-items: center;
+    text-align: center;
```
lfittl, line 93: This was necessary since the install overview page in the public docs unfortunately now spills into two lines for Azure. It doesn't look great, but I think this is okay-ish, and the in-app version does not suffer from this problem since its wider (see screenshot in PR description).

Screenshot 2023-09-29 at 10 54 02 AM

msakrejda (approved):

PR #190, opened 2023-10-04
Document the new db_data_directory setting

msakrejda:


collector/settings.mdx
```diff
@@ -174,6 +174,11 @@ helper functions are expected to be defined, and the one we connect to for serve
       <td>[none]</td>
       <td>Alternative to above, using actual contents of the key</td>
     </tr>
+    <tr>
+      <td>db_data_directory (<code>DB_DATA_DIRECTORY</code>)</td>
+      <td>[autodetected]</td>
```
lfittl, line 179: It appears we use "auto-detected" without the brackets for other settings (see e.g. aws_db_instance_id), might be worth standardizing on one style.

collector/settings.mdx
```diff
@@ -174,6 +174,11 @@ helper functions are expected to be defined, and the one we connect to for serve
       <td>[none]</td>
       <td>Alternative to above, using actual contents of the key</td>
     </tr>
```
lfittl, line 176: I wonder if this wouldn't be a better fit for the "Self-managed servers" section, since its only relevant in that case, and `db_log_location` also lives there.

lfittl (approved):

PR #196, opened 2023-10-25
Improve Xmin horizon check details

lfittl:
```
Xmin horizon check: Show TXID with separate epoch, and assigned at time

Since the statistics views in Postgres utilize the 32-bit version of
transaction IDs, but we track 64-bit internally, it can be confusing
to show the raw 64-bit value (since you can't visually compare it to
queries you run on the database). To improve, instead format it with
a separate epoch, separated by a colon, like we do for freezing stats.

Additionally show the specific assigned at timestamp inline, to help
identify the problem origin better, e.g. to use for comparison with
xact_start in pg_stat_activity, when xmin horizon is held back by a
long-running transaction.

In passing introduce a component to handle xmin information consistently,
and fix typing for "heldBackInfo", which was incorrectly marked as a raw
number (its an object).
```
```
Xmin horizon check: Clarify that pg_terminate_backend is needed sometimes

To stop a long-running transaction that's idle, one needs to terminate
the connection, not just cancel (the non-existent) active query.

Additionally add a stronger disclaimer about the possible impact of such
an action.
```

Before:

Screenshot 2023-10-25 at 12 04 12 AM

After:

Screenshot 2023-10-25 at 12 05 33 AM

### DIFF
--- a/components/CheckDocumentation/vacuum/XminHorizon.tsx
+++ b/components/CheckDocumentation/vacuum/XminHorizon.tsx
@@ -129,18 +129,11 @@ function formatFullTransactionID(value: number): string {
   return `${epoch}:${xid}`;
 }
 
-const XminHeldBackInfo: React.FunctionComponent<{
-  type: string;
+const XminHeldBackPoint: React.FunctionComponent<{
   info: HeldBackInfoType;
-}> = ({ type, info }) => {
+}> = ({ info }) => {
   const fullTxid = formatFullTransactionID(info["xmin"]);
-
-  return (
-    <>
-      A {type} is holding back the xmin horizon at{" "}
-      <code>{`${fullTxid}`}</code> (assigned at {new Date(info["assigned_at"] * 1000).toISOString()})
-    </>
-  );
+  return <code>{`${fullTxid}`}</code>;
 }
 
 const GuidanceByBackend: React.FunctionComponent<{
@@ -164,7 +157,8 @@ const GuidanceByBackend: React.FunctionComponent<{
       </h6>
       {heldBackInfo && (
         <p>
-          <XminHeldBackInfo type="long-running transaction" info={heldBackInfo} />.
+          A long-running transaction is holding back the xmin horizon at{" "}
+          <XminHeldBackPoint info={heldBackInfo} />.
         </p>
       )}
       <p>
@@ -187,7 +181,7 @@ const GuidanceByBackend: React.FunctionComponent<{
       <CodeBlock>
         <SQL sql={`SELECT pg_terminate_backend('<query_pid>');`} />
       </CodeBlock>
-      <p>Note this will roll back the transaction, and <strong>discard all data written to the database earlier within that transaction</strong>.</p>
+      <p>Note this will roll back the transaction, and <strong>discard all data</strong> written to the database earlier within that transaction.</p>
     </li>
   );
 };
@@ -218,7 +212,8 @@ const GuidanceByReplicationSlot: React.FunctionComponent<{
       </h6>
       {heldBackInfo && (
         <p>
-          <XminHeldBackInfo type="replication slot" info={heldBackInfo} />.
+          A replication slot is holding back the xmin horizon at{" "}
+          <XminHeldBackPoint info={heldBackInfo} />.
         </p>
       )}
       <p>
@@ -263,7 +258,8 @@ const GuidanceByReplicationSlotCatalog: React.FunctionComponent<{
       </h6>
       {heldBackInfo && (
         <p>
-          <XminHeldBackInfo type="replication slot" info={heldBackInfo} />,
+          A replication slot is holding back the xmin horizon at{" "}
+          <XminHeldBackPoint info={heldBackInfo} />,
           specifically with system catalogs.
         </p>
       )}
@@ -307,7 +303,8 @@ const GuidanceByStandby: React.FunctionComponent<{
       </h6>
       {heldBackInfo && (
         <p>
-          <XminHeldBackInfo type="long running query on a standby" info={heldBackInfo} />.
+          A long running query on a standby is holding back the xmin horizon at{" "}
+          <XminHeldBackPoint info={heldBackInfo} />.
         </p>
       )}
       <p>
@@ -364,7 +361,8 @@ const GuidanceByPreparedXact: React.FunctionComponent<{
       </h6>
       {heldBackInfo && (
         <p>
-          <XminHeldBackInfo type="prepared transaction" info={heldBackInfo} />.
+          A prepared transaction is holding back the xmin horizon at{" "}
+          <XminHeldBackPoint info={heldBackInfo} />.
         </p>
       )}
       <p>
--- a/components/CheckDocumentation/vacuum/XminHorizon.tsx
+++ b/components/CheckDocumentation/vacuum/XminHorizon.tsx
@@ -8,6 +8,7 @@ import {
 } from "../../../util/checks";
 import { useCodeBlock } from "../../CodeBlock";
 import { useSmartAnchor } from "../../SmartAnchor";
+import PGDocsLink from "../../PGDocsLink";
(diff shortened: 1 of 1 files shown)


components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -98,9 +98,48 @@ const XminHorizonGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
...
+  const epoch = epochFromFullTransactionId(value);
+  return fullTransactionIdMinusEpoch(value, epoch);
+}
+
+function fullTransactionIdMinusEpoch(
+  value: number,
+  epoch: number,
+): number {
+  return Number(BigInt(value) - (BigInt(epoch) << BigInt(32)));
+}
```
lfittl, line 123: These functions are from `app/javascript/utils/txid.ts` in the main app, but I don't see a good way to share these (since this is also included in the public website code), thus copied them.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -135,17 +173,22 @@ const GuidanceByBackend: React.FunctionComponent<{
                   ORDER BY greatest(age(backend_xmin), age(backend_xid)) DESC;`}
         />
       </CodeBlock>
-      <p>You can cancel it by running either of commands:</p>
+      <p>You can cancel it by running either <a href="https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SIGNAL" target="_blank">pg_cancel_backend</a> if the transaction is running an active query:</p>
```
msakrejda, line 176: We have `<PGDocsLink />` for linking to Postgres documentation; we should probably use that here.

components/CheckDocumentation/vacuum/XminHorizon.tsx
msakrejda: Given the limited amount of code and the fact that it won't change much, I think this is okay. Maybe we should add a comment referencing the source? If you think it's not appropriate to reference the private code in the public docs, I'm okay to leave as-is.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -98,9 +98,48 @@ const XminHorizonGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
...
+  type: string;
+  info: HeldBackInfoType;
+}> = ({ type, info }) => {
+  const epoch = epochFromFullTransactionId(info["xmin"]);
+  const xid = xidFromFullTransactionId(info["xmin"]) as number;
+
+  return (
+    <>
+      A {type} is holding back the xmin horizon at{" "}
+      <code>{`${epoch}:${xid}`}</code> (assigned at {new Date(info["assigned_at"] * 1000).toISOString()})
```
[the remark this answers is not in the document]
msakrejda: No strong feelings. Though if you keep it, maybe this expression could be pulled up to a local in the component to keep the JSX focused on rendering.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -98,9 +98,48 @@ const XminHorizonGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
...
+): number {
+  return Number(BigInt(value) - (BigInt(epoch) << BigInt(32)));
+}
+
+const XminHeldBackInfo: React.FunctionComponent<{
+  type: string;
+  info: HeldBackInfoType;
+}> = ({ type, info }) => {
+  const epoch = epochFromFullTransactionId(info["xmin"]);
+  const xid = xidFromFullTransactionId(info["xmin"]) as number;
```
msakrejda, line 130: Why `as number`?

components/CheckDocumentation/vacuum/XminHorizon.tsx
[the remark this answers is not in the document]
msakrejda: Reading through this again, I think the way it composes with its JSX "callers" is kind of weird. I think this would work better as just something like

```tsx
const XminHeldBackPoint: React.FunctionComponent<{
  info: HeldBackInfoType;
}> = ({ type, info }) => {
  const epoch = epochFromFullTransactionId(info["xmin"]);
  const xid = xidFromFullTransactionId(info["xmin"]) as number;
  const assignedAt = new Date(info["assigned_at"] * 1000).toISOString();
  return (
    <>
      <code>{`${epoch}:${xid}`}</code> (assigned at {assignedAt})
    </>
  )
}
```

at the cost of a bit of duplication of the "holding back the xmin horizon" text. It'd be easier to tell what the actual messages would be. Not a super-strong feeling, though.

msakrejda (approved):
Some minor comments, but looks good.

components/CheckDocumentation/vacuum/XminHorizon.tsx
keiko713: I was going to suggest to add `xmin_txt` or something as a new hash value so that xmin will keep xid8 whereas `xmin_txt` will have `${epoch}:${xid}` (then we don't need to add a duplicated xid logic here).
Then, I sadly realized that we actaully don't have the logic to make `xmin_txt` in our backend handy (it's not so difficult to get `xid` part, we could do `xmin_horizon_backend::text::xid` during SELECT, but I don't think there is a handy way to get an epoch) so this code needs to be added somewhere newly :/
With that, I'd say that it's okay to add it here. Re: comment, how about adding it in the private repo side?

components/CheckDocumentation/vacuum/XminHorizon.tsx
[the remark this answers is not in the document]
keiko713: I was feeling a bit strange to see that "assigned at" in screenshot (thanks for adding them!), the format is not really something we use elsewhere, also as you mentioned, that info should be in the page already. I'd vote for dropping.

msakrejda:
Currently waiting on https://github.com/pganalyze/pganalyze/pull/4773

components/CheckDocumentation/vacuum/XminHorizon.tsx
msakrejda: To answer my own question, this is probably because `xidFromFullTransactionId` can return null if passed null. It never is here, but there's no point in updating the function and diverging from the original implementation.

components/CheckDocumentation/vacuum/XminHorizon.tsx
```diff
@@ -135,17 +173,22 @@ const GuidanceByBackend: React.FunctionComponent<{
...
-      <p>You can cancel it by running either of commands:</p>
+      <p>You can cancel it by running either <a href="https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SIGNAL" target="_blank">pg_cancel_backend</a> if the transaction is running an active query:</p>
       <CodeBlock>
         <SQL sql={`SELECT pg_cancel_backend('<query_pid>');`} />
       </CodeBlock>
+      <p>or <a href="https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SIGNAL" target="_blank">pg_terminate_backend</a> if the connection state is "idle in transaction".</p>
+      <CodeBlock>
+        <SQL sql={`SELECT pg_terminate_backend('<query_pid>');`} />
+      </CodeBlock>
+      <p>Note this will roll back the transaction, and <strong>discard all data written to the database earlier within that transaction</strong>.</p>
```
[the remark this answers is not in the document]
msakrejda: I decided to keep the `<strong>`, but narrow it to "discard all data". That still highlights the danger without belaboring the point.

msakrejda:
Okay, I rebased and addressed review feedback (including mine :grin:). Ready for review.

lfittl (commented):
:+1: Still looks good to me, thanks for updating it!

PR #198, opened 2023-11-07
Add description for DB_ALL_NAMES env var for collector settings

keiko713:
Just realized that `DB_ALL_NAMES` is not documented, so adding it. It's introduced back in 2017:

https://github.com/pganalyze/collector/blob/b366c7fa7c152894758159602cb599fa2412760a/CHANGELOG.md?plain=1#L1363

### DIFF
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -174,6 +174,14 @@ helper functions are expected to be defined, and the one we connect to for serve
       <td>[none]</td>
       <td>Alternative to above, using actual contents of the key</td>
     </tr>
+    <tr>
+      <td><code>DB_ALL_NAMES</code></td>
+      <td>false</td>
+      <td>Alternative to setting <code>*</code> at the last entry of db_name (<code>DB_NAME</code>),
+      monitoring all databases on server. Can be only set using an environment variable.<br />
+      This is helpful when db_url (<code>DB_URL</code>) is used to specify a monitoring server (db_name is unused)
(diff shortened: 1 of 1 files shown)


seanlinsley (approved):

collector/settings.mdx
```diff
@@ -174,6 +174,14 @@ helper functions are expected to be defined, and the one we connect to for serve
       <td>[none]</td>
       <td>Alternative to above, using actual contents of the key</td>
     </tr>
+    <tr>
+      <td><code>DB_ALL_NAMES</code></td>
+      <td>false</td>
+      <td>Alternative to setting <code>*</code> at the last entry of db_name (<code>DB_NAME</code>),
+      monitoring all databases on server. Can be only set using an environment variable.<br />
```
lfittl, line 181: "Can be only set" => "Can only be set"

Paging @msakrejda to confirm that this rewording would be more correct.

msakrejda, line 181: +1 on the updated wording, and a few more suggestions:

```suggestion
      <td>Alternative to setting <code>*</code> as the last entry of db_name (<code>DB_NAME</code>) to
      monitor all databases on the server. Can only be set using an environment variable.<br />
```

I think we left this undocumented intentionally because the wildcard syntax came later, but the use case you describe below seems reasonable, so no objection to adding this to the docs.

msakrejda (approved):

PR #199, opened 2023-11-17
Remove changelog page of Indexing Engine

keiko713:
As discussed internally, we decided to remove this section.


lfittl (approved):
For additional context to anyone external following this: We have work under way on a public changelog for all parts of the product, better integrated with our development workflow, that will include any Indexing Engine changes.

Since this changelog here has become stale (since it was manually maintained), we decided to remove it for now for clarity.

PR #200, opened 2023-11-21
Add docs for early access constraint programming model

msepga:
This PR includes the constraint programming model docs with some changes:

- Split up document into multiple pages
- Fix links to point to new sub-pages
- Replace absolute links with relative links (when referencing e.g. the "good enough" model)


lfittl:
@msepga Nice! I've done a further revision/simplification on this in an additional commit I just pushed - will merge as a first version, and we can refine further as we see fit.

lfittl (approved):

PR #201, opened 2023-11-29
Update docs for updated collector output

msakrejda:
See https://github.com/pganalyze/collector/pull/482


lfittl (approved):
Excellent, thank you for making the adjustments! (and thinking of doing so)

PR #202, opened 2023-11-30
Update Crunchy Bridge EXPLAIN integration to use auto_explain

keiko713:
Update Crunchy Bridge EXPLAIN integration docs to use auto_explain instead of Log-based EXPLAIN.
You can check out the part that is newly added in below.

https://pganalyze-we-crunchy-au-hqjvid.herokuapp.com/docs/explain/setup/crunchy_bridge/01_auto_explain_check

Step 2 doesn't load well (loads AWS one) for some reason to me, but it loads okay locally so it's something to do with cache?


explain/setup/crunchy_bridge/_01_in_app.mdx
```diff
@@ -0,0 +1,34 @@
...
+      </Link>
+    </React.Fragment>
+  )
+}
+
+export const AutoExplainCheck = ({enabled}) => {
+  return enabled ? <AutoExplainEnabled /> : <AutoExplainNotEnabled />
+}
+
+<AutoExplainCheck enabled={props.autoExplainEnabled} />
```
keiko713, line 34: note: this doesn't return `true` when it's enabled with `session_preload_libraries` currently, though I'll make a PR to support it too in app, so let's assume that it's doing its job in this PR.

msakrejda (commented):
We've talked about documenting this in the past, and I think there were some concerns around recommending `session_preload_libraries`, I think around performance.

lfittl:
> We've talked about documenting this in the past, and I think there were some concerns around recommending `session_preload_libraries`, I think around performance.

I tried to remember what the concern was and couldn't recall. It seems like Crunchy documents this publicly now, so I think its probably acceptable to do so as well: https://docs.crunchybridge.com/extensions-and-languages/auto_explain

msakrejda:
Makes sense. I checked to see if we should update Aiven since they had similar restrictions, but the [state of the art there](https://docs.aiven.io/docs/products/postgresql/concepts/dba-tasks-pg) still seems to be wrapper functions, which I think are complicated enough to not bother documenting on our side (I think interested users can figure it out).

explain/setup/crunchy_bridge/03_review_settings.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+There are a number of `auto_explain` settings that allow you to configure which queries
+will generate EXPLAIN plans. You can read the full Postgres documentation about
+these <AutoExplainLink version={props.pgMajorVersion}>here</AutoExplainLink>.
+The right configuration will depend on your server and workload,
+but we've found the following is a good starting point:
+
+<PgSettingsRecommendations mode='alter system' recommendations={getAutoExplainSettingRecommendations(props.settings)} />
+
+Run the above commands as a Postgres superuser. Then, to apply these settings, run
```
msakrejda, line 26: Scanning the Crunchy docs, it looks like they do allow true superuser, right?

lfittl: Yep, they do.

PR #205, opened 2023-12-07
Clarify how to specify the managed identity when multiple

keiko713:
Pairing PR of https://github.com/pganalyze/collector/pull/486
Technically blocked by that PR, but given that it's actually not working currently, I think it's safe to ship this independently too.

With above collector change, following the documentation should complete the setup without a problem. However, there is a caveat when multiple identities are assigned to the VM, so let's clarify what needs to be done in that case.

### DIFF
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -410,7 +410,11 @@ Only relevant if you are running your database in Azure using Azure Database for
     <tr>
       <td>azure_ad_client_id (<code>AZURE_AD_CLIENT_ID</code>)</td>
       <td>[none]</td>
-      <td>The "Application (client) ID" on your application. Alternatively, when using a Managed Identity, this can also be used to define which identity to use, when a VM has multiple identities assigned.</td>
+      <td>
+        The "Application (client) ID" on your application.
+        For Managed Identity, particularly with multiple VM identities, avoid using this parameter.
+        Instead, set the <code>AZURE_CLIENT_ID</code> environment variable to specify the identity
+      </td>
     </tr>
     <tr>
       <td>azure_ad_client_secret (<code>AZURE_AD_CLIENT_SECRET</code>)</td>
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -410,11 +410,7 @@ Only relevant if you are running your database in Azure using Azure Database for
     <tr>
       <td>azure_ad_client_id (<code>AZURE_AD_CLIENT_ID</code>)</td>
       <td>[none]</td>
-      <td>
-        The "Application (client) ID" on your application.
-        For Managed Identity, particularly with multiple VM identities, avoid using this parameter.
-        Instead, set the <code>AZURE_CLIENT_ID</code> environment variable to specify the identity
-      </td>
+      <td>The "Application (client) ID" on your application. Alternatively, when using a Managed Identity, this can also be used to define which identity to use, when a VM has multiple identities assigned.</td>
     </tr>
     <tr>
       <td>azure_ad_client_secret (<code>AZURE_AD_CLIENT_SECRET</code>)</td>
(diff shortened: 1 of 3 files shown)


msakrejda (approved):

lfittl (commented):
Per my comment on https://github.com/pganalyze/collector/pull/486, I do think that we should just go ahead and move from "AZURE_AD_" to "AZURE_" as the prefix for all the settings, as that matches how you usually configure Azure credentials (see e.g. https://learn.microsoft.com/en-us/dotnet/api/azure.identity.environmentcredential?view=azure-dotnet).

I don't think there is a reason to keep the "AZURE_AD_" prefix when using the static credentials via an application.

log-insights/setup/azure-database/05_configure_collector.mdx
```diff
@@ -18,6 +18,7 @@ Next, update the collector configuration for this server. Note the **values need
 ]} configFromEnv={props.configFromEnv} />
 
 Thanks to the Managed Identity that's assigned to the VM, we don't need to specify any credentials here.
+If the VM has multiple Managed Identities assigned, specify the client ID of the specific Managed Identity created in Step 1 using `AZURE_CLIENT_ID` environment variable.
```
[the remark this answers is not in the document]
keiko713: I added "the", though changed here a bit to use "the `azure_ad_client_id` / `AZURE_AD_CLIENT_ID` setting" to match to our troubleshooting doc

collector/settings.mdx
```diff
@@ -410,7 +410,11 @@ Only relevant if you are running your database in Azure using Azure Database for
     <tr>
       <td>azure_ad_client_id (<code>AZURE_AD_CLIENT_ID</code>)</td>
       <td>[none]</td>
-      <td>The "Application (client) ID" on your application. Alternatively, when using a Managed Identity, this can also be used to define which identity to use, when a VM has multiple identities assigned.</td>
+      <td>
+        The "Application (client) ID" on your application.
+        For Managed Identity, particularly with multiple VM identities, avoid using this parameter.
+        Instead, set the <code>AZURE_CLIENT_ID</code> environment variable to specify the identity
```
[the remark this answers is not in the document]
keiko713: Per https://github.com/pganalyze/collector/pull/486, I decided to keep the environment variable, therefore the change of this part is not needed anymore, so I dropped the change

log-insights/setup/azure-database/troubleshooting.mdx
```diff
@@ -32,6 +32,7 @@ When using Managed Identity, verify that you have actually assigned the identity
 This may also occur when you have multiple user-assigned managed identities set for your virtual machine.
 Try removing any unrelated managed identities, or explicitly set the `azure_ad_client_id` / `AZURE_AD_CLIENT_ID` setting to
 the Managed Identity's Client ID.
+You may also observe the error message like "ManagedIdentityCredential: no default identity is assigned to this resource" in this case.
```
keiko713, line 35: This part is talking about JWT, which we already retired with 0.50.0 but I am lazy to rewrite everything, so I just added a new error message.
The lazy alternative is to update the title from `Error: "failed to configure Azure AD JWT provider"` to `Error: "ManagedIdentityCredential: no default identity is assigned to this resource"` here.
Thoughts?

keiko713:
memo: merge this after the collector new release

log-insights/setup/azure-database/troubleshooting.mdx
msakrejda: While I think the docs should focus on the latest collector (since this information is most relevant for initial setup, and you're most likely to be setting up a collector right after downloading the latest version), it's helpful to keep troubleshooting information for users of collectors at least a few versions back. They may need to change their configuration and decide to do so without updating the collector and run into problems. I also think it's helpful to "key" troubleshooting information by the error you're seeing.

Given that, is this just a new error message for the same troubleshooting steps? If so, I would change the section title, and add a new section with the old title that just says

> See the above section titled Error: "ManagedIdentityCredential: no default identity is assigned to this resource"

keiko713: Nice, I like that idea. Will update that way 👍

PR #207, opened 2023-12-07
Update example log line for new collector output format

msakrejda:
This was missed in #201.


lfittl (approved):
Good catch - I think that one is on me for forgetting to update it here.

msakrejda:
I had it an unsaved tab in my editor: I think I just neglected to save before submitting #201.

PR #210, opened 2023-12-12
Improve docs for receiving syslog over TLS

keiko713:
The collector side change: https://github.com/pganalyze/collector/pull/487

* Updated collector settings page to mention new config values
* Updated Aiven log insights setup docs to mention that using TLS is also an option
* Updated Syslog server setup page to mention that using TLS is also an option


keiko713:
Ugh totally forgot to update Aiven's docs, it's talking about "you need to move database into vpc to enable Log Insights":

https://pganalyze.com/docs/log-insights/setup/aiven/01_move_database_into_vpc

We should update this to say that the other way could be used, I'll make an update tomorrow.

keiko713:
Updated, ready for review (we should merge after the collector release though)

log-insights/setup/aiven/01_move_database_into_vpc.mdx
```diff
@@ -12,11 +12,17 @@ export const ImgAivenCreateVpc = () => <div style={{display: 'flex', justifyCont
...
-running in. In later steps we refer to this VPC as the "collector VPC". If you
-have not done so yet, go to the Aiven console and create a VPC for the project
-in the same region where your database is running:
+In order to securely deliver logs to the collector without enabling TLS, your
+database must run in an Aiven VPC peered to the VPC of the EC2 instance or
+container the collector is running in.
+Alternatively, you can send logs over TLS by setting up certificates for the
+collector syslog server. See [Receiving syslog over TLS](https://pganalyze.com/docs/log-insights/setup/syslog-server#receiving-syslog-over-tls)
+for how to set up certificates. Once it's done, go to <Link to="04_configure_service_integration">Step 4</Link>
+to configure Service Integration.
```
msakrejda, line 21: As is, the TLS syslog option is sandwiched in the middle of the description of the peering option. Maybe since the TLS option is simpler, we should talk about that first in its own paragraph, and then go into the peering instructions? E.g., something like

> In order to securely deliver logs to the collector, there are two possible approaches: either configure syslog to work over TLS by setting up certificates for the collector syslog server, or use VPC peering between your database and the collector.
>
> In order to set up syslog over TLS, see \[Receiving syslog over TLS\](https://pganalyze.com/docs/log-insights/setup/syslog-server#receiving-syslog-over-tls) for how to set up certificates, then continue to \<Link to="04_configure_service_integration"\>Step 4\</Link\>. To set up VPC peering, continue below.
>
> To use VPC peering, your database must run in an Aiven VPC peered to the VPC of the EC2 instance or container the collector is running in. In later steps we refer to this VPC as the "collector VPC"....

I think ideally, we'd actually have separate break-out sections for these two options (like we have for, e.g., EC2 instance versus ECS), but that's a headache and I don't think it's worth doing right now.

log-insights/setup/aiven/04_configure_service_integration.mdx
```diff
@@ -25,14 +25,25 @@ The last step is setting up the Aiven Service Integration. Go to the
 Service Integration page in your Aiven console, select Syslog from the
 list of options, and click "Add new endpoint".
 
+<strong>Setup using VPC</strong>
```
msakrejda, line 28: Not a big deal, but can we use a `### subheading` here? (Possible with more or less `#`s.)

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
 ---
 
 If your database logs are available via syslog, you can use the [`db_log_syslog_server`](/docs/collector/settings#self-managed-servers)
-setting to have the collector listen for syslog messages from a remote server. This should be a local address like
-"0.0.0.0:32514" and it's recommended to use an unprivileged port number (1024 and up) to avoid running
-the collector as root.
+setting to have the collector listen for syslog messages from a remote server.
+When this is specified, the collector will run the syslog server behind the
+scene on the specified port to receive syslog messages.
```
msakrejda, line 10:
```suggestion
scenes on the specified port to receive syslog messages.
```

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
...
+This should be a local address like "0.0.0.0:32514" and it's recommended to use
+an unprivileged port number (1024 and up) to avoid running the syslog server as
+root.
 
-Note that you need to use protocol RFC5424, with an unencrypted TCP connection. Due to syslog not being
-an authenticated protocol it is recommended to only use this integration over private networks.
+Note that you need to use protocol RFC5424, with a TCP connection. When not
+using TLS, it is recommended to only use this integration over private networks.
```
msakrejda, line 16: I feel like this waters down this warning a bit. If you're using TLS but _not_ using client certs, someone could still be sending bogus logs to your collector and you won't be able to tell. Now, the likelihood of this happening is pretty small, and the damage they can do is pretty limited (DoS your collector? cause bogus information to show up in your Log Insights page?), but I think I'd still rather make it clear what the situation is and allow the user to decide whether they want to take that chance.

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
...
+
+### Receiving syslog over TLS
+
+You can also receive logs over TLS by specifying the path of a certificate [`db_log_syslog_server_cert_file`](/docs/collector/settings#self-managed-servers)
+and a secret key [`db_log_syslog_server_key_file`](/docs/collector/settings#self-managed-servers).
+Alternatively, you can also set the actual contents of a certificate with [`db_log_syslog_server_cert_contents`](/docs/collector/settings#self-managed-servers),
+or a secret key with [`db_log_syslog_server_key_contents`](/docs/collector/settings#self-managed-servers).
+Please note that if the `db_log_syslog_server_cert_file` setting is provided,
+the connection needs to be made as TLS and the connection attempt without TLS
```
msakrejda, line 25:
```suggestion
the connection needs to be made as TLS and connection attempts without TLS
```

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
...
+will be rejected.
+
+This certificate is for the syslog server running on the collector. You have the
```
msakrejda, line 28:
```suggestion
This certificate is for the syslog server running as part of the collector. You have the
```

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
...
+option to use an existing certificate or generate a new self-signed certificate
+specifically for this purpose. In cases where a self-signed certificate or a
+certificate from an untrusted authority is used, it is also essential to specify
+the path to the TLS Certificate Authority (CA) certificate(s) with the [`db_log_syslog_server_ca_file`](/docs/collector/settings#self-managed-servers)
+setting. This CA certificate is also required to provide to the client that is
+sending out logs to the syslog server, such as rsyslog.
```
msakrejda, line 34:
```suggestion
setting. This CA certificate must also be provided to the client that is
sending out logs to the syslog server, such as rsyslog.
```

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
...
+
+With these settings, logs from the client (like rsyslog) to the collector syslog
+server will be sent over TLS.
+
+You can also enable mutual TLS to ensure that the client side is also verified.
+The client side can provide a certificate for this, and if this certificate is
+a self-signed certificate or a certificate from an untrusted authority, the
+collector syslog server needs to know about the CA certificate of it. You can
```
msakrejda, line 42:
```suggestion
collector syslog server needs to know about the CA certificate for it. You can
```

log-insights/setup/syslog-server.mdx
```diff
@@ -5,9 +5,40 @@ backlink_title: 'Log Insights: Setup'
...
+specify the CA certificate of the client side using the [`db_log_syslog_server_client_ca_file`](/docs/collector/settings#self-managed-servers)
+setting.
```
msakrejda, line 44:
```suggestion
or [`db_log_syslog_server_client_ca_contents`](/docs/collector/settings#self-managed-servers) setting.
```

msakrejda (approved):

log-insights/setup/aiven/04_configure_service_integration.mdx
keiko713: I found that these also-shown-in-app markdown files are not using markdown that much for some reason (well some of them are legit reasons, like images and links), but I think `###` is supported. I'll change this and see how it looks in both web and in app.

log-insights/setup/syslog-server.mdx
keiko713: Yeah fair point. I'll tweak to cover that

keiko713: Uhmm while trying to tweak this, I started wondering about our choice `VerifyClientCertIfGiven` https://pkg.go.dev/crypto/tls#ClientAuthType

You're right that the using TLS won't be totally secure, in terms of anyone can send out logs to the collector (still secure in terms of eavesdropping and tampering). When someone actually wants to ensure the mTLS (aka prevent from any random folks sending out logs), they can't really do this with `VerifyClientCertIfGiven`, because we only verify when the client cert is given, so bad person can still send out logs without the client cert and the collector will just receive it.
I think `VerifyClientCertIfGiven` is still valid as a default behavior, though we might consider supporting additional options if there's a request for such a feature.

log-insights/setup/aiven/04_configure_service_integration.mdx
keiko713: yup, `###` worked fine:

In web:

image

In App:

image

log-insights/setup/syslog-server.mdx
keiko713: Update, maybe I'd love to have one more look this part (no rush).

keiko713: Gonna merge as is but feel free to leave any comments if you have, happy to create a follow-up PR

msakrejda: The revised text looks good. Good point on `VerifyClientCertIfGiven`. I think you're right; we can address that if it comes up.

PR #213, opened 2024-01-24
Index Advisor: Missing Indexes issue docs

msakrejda:
Add a Missing Indexes issue. This is like the existing Missing Index
issue, but it looks at all index recommendations on a single table
together, rather than recommending individual indexes in separate
issues.

### DIFF
--- a/components/CheckDocumentation/index.tsx
+++ b/components/CheckDocumentation/index.tsx
@@ -7,7 +7,7 @@ import ActiveQuery from "./connections/ActiveQuery";
 import IdleTransaction from "./connections/IdleTransaction";
 import BlockingQuery from "./connections/BlockingQuery";
 import MissingIndex from "./index_advisor/MissingIndex";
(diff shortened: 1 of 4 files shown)


components/CheckDocumentation/index.tsx
```diff
@@ -36,6 +37,7 @@ const Docs: {
   },
   index_advisor: {
     missing_index: MissingIndex,
+    index_insights: IndexInsights,
```
msakrejda, line 40: Note that internally this is named `IndexInsights` / `index_insights`, since future work will enable insights that recommend dropping indexes. Since changing some of these internal names is harder, we want to use a more generic name there right off the bat.

keiko713 (approved):
Are we eventually removing "missing_index" check from here?

msakrejda:
@keiko713 I think that's the plan, yeah. Having all index recommendations on a single table together should be strictly superior.

PR #214, opened 2024-02-08
Add docs pages for Schema Statistics, column stats and HOT updates

lfittl:


schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: 'Column statistics and HOT updates'
+backlink_href: /docs/schema-statistics
+backlink_title: 'Tracking Postgres schema statistics in pganalyze'
+---
+
+![](table-columns.png)
+
+On the "Columns" tab of each table you can find all columns that are defined on the table, as well as the planner statistics associated with the column (collected by ANALYZE):
```
keiko713, line 9:
```suggestion
On the "Columns" tab of each table, you can find all columns that are defined on the table, as well as the planner statistics associated with the column (collected by ANALYZE):
```

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+* **Null %**: Fraction of the rows in this table that have a NULL value in this column.
+* **Avg. Size**: Average size, in bytes, of the values in this column across all the rows in this table (excluding NULLs).
+* **Num. Distinct**: Estimated number of distinct values in this column. If followed by an asterisk (*), the number of values is expected to grow as the table grows (e.g., for a unique value like a primary key).
```
msakrejda, line 13: This is from the app copy, right? Looking at it again, I think we should emphasize "distinct" here:

```suggestion
* **Num. Distinct**: Estimated number of distinct values in this column. If followed by an asterisk (*), the number of distinct values is expected to grow as the table grows (e.g., for a unique value like a primary key).
```

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+In case of partitioned tables you will additionally see "Full NULL %", "Full Avg Size" and a second "Num. Distinct" data point. This represents the data that Postgres collects as an aggregate information across all partitions, as compared to just the partitioned table itself (in case of inheritance-based partitioning).
```
keiko713, line 15:
```suggestion
In case of partitioned tables, you will additionally see "Full NULL %", "Full Avg Size" and a second "Num. Distinct" data point. This represents the data that Postgres collects as an aggregate information across all partitions, as compared to just the partitioned table itself (in case of inheritance-based partitioning).
```

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+Note that planner statistics are only available when the [get_column_stats](/docs/install/troubleshooting/column_stats_helper) helper is set up on your database.
+
+## Heap-Only Tuples (HOT) optimization
+
+On the "Columns" tab you can also see pganalyze's assessment whether UPDATEs on a given column can use the [Heap-Only Tuples (HOT)](https://www.postgresql.org/docs/current/storage-hot.html) optimization:
```
keiko713, line 21:
```suggestion
On the "Columns" tab, you can also see pganalyze's assessment whether UPDATEs on a given column can use the [Heap-Only Tuples (HOT)](https://www.postgresql.org/docs/current/storage-hot.html) optimization:
```

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+* **Updates / Min**: How often this column is updated per minute on average, based on UPDATE statements and CTEs containing UPDATEs on this table. If multiple statements update this column, this is the sum of the average number of calls. Excludes any queries for which query analysis failed (see Index Advisor status).
```
keiko713, line 23: What exactly is "Index Advisor status" here?

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+* **Indexed?**: Indicates whether there are any indexes that include this column. This also considers indexes that have expressions referencing the column, as well as partial and covering indexes.
+* **HOT?**: Indicates whether updates on this column (if any) can create Heap Only Tuples (HOT). Generally, updates are not HOT capable when the column is indexed. For best performance and to reduce VACUUM overhead, avoid creating indexes on columns that are updated frequently.
```
keiko713, line 25:
```suggestion
* **HOT?**: Indicates whether updates on this column (if any) can create Heap-Only Tuples (HOT). Generally, updates are not HOT capable when the column is indexed. For best performance and to reduce VACUUM overhead, avoid creating indexes on columns that are updated frequently.
```

schema-statistics/columns-hot-updates.mdx
msakrejda, line 15:
```suggestion
In case of partitioned tables you will additionally see "Full NULL %", "Full Avg. Size" and a second "Num. Distinct" data point. This represents the data that Postgres collects as aggregate information across all partitions, as compared to just the partitioned table itself (in case of inheritance-based partitioning).
```

Also, this is for inheritance, not partitioning right? The most common usage is probably still partitioning, but I think we should focus on the underlying mechanism, since that's what the stats refer to. So maybe:

```suggestion
In case of tables with inheritance children (e.g., if using the older inheritance-based partitioning mechanism), you will additionally see "Full NULL %", "Full Avg. Size" and a second "Num. Distinct" data point. This represents the data that Postgres collects as aggregate information across all children (i.e., individual partitions), as compared to just the table itself.
```

We could maybe even add a "Note that this not apply to the new declarative partitioning mechanism (since the parent partitioned table is always empty). Check the [Postgres documentation on partitioning](https://www.postgresql.org/docs/current/ddl-partitioning.html) for details."

schema-statistics/columns-hot-updates.mdx
msakrejda, line 23: Is this average over the selected time period, or an arbitrary time period? Should we mention this (on the fence here; that does complicate the docs).

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+HOT Updates generally have a significantly lower overhead, since they do not require any index updates. They also enable on-access HOT pruning, which reduces the effort that a future autovacuum will have to do to clean up dead rows.
```
keiko713, line 27: "HOT Updates", are these updates that are HOT capable? Should it be HOT UPDATEs (though feels a bit off), or maybe "HOT capable UPDATEs" 🤔

schema-statistics/index.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+![](table-treemap.png)
+
+In the schema statistics overview you can find two treemaps, as well as full lists of all the tables and indexes on a given database. The treemap is useful to identify how a specific index or table size relates to the total.
+
+## Table growth over time
+
+![](table-growth.png)
+
+For each table you can track growth over time, understand how a table utilizes TOAST (The Oversized Attribute Storage Technique) for storing large values, and see the aggregate size for all indexes on the table.
```
msakrejda, line 19: Maybe link to https://www.postgresql.org/docs/current/storage-toast.html ?

schema-statistics/columns-hot-updates.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+The HOT information is also utilized for calculating "Update Overhead" in the [constraint-programming model](/docs/indexing-engine/cp-model) for pganalyze Index Advisor (currently in early access), to avoid indexing columns that are being updated.
```
keiko713, line 29: I don't have a better alternative but "currently in early access" could mislead that the Index Advisor itself is in early access (which is not). Maybe... we could remove it (so that we don't need to remember removing that part when launching the new version)?

schema-statistics/index.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+![](index-details.png)
+
+For each individual index you can track usage patterns over time, whether the index scans were able to use the Postgres shared buffer cache, as well as the date the index was last scanned. You can receive alerts for unused indexes using the [Unused Indexes check](/docs/checks/schema/index_unused).
+
+## Queries and scans per table
+
+![](table-queries-and-scans.png)
+
+As queries are received by pganalyze, they are parsed, and an internal association is stored that allows looking up query activity by table. Additionally, each query is broken down into the [individual scans](/docs/indexing-engine/query-to-scan-breakdown).
```
msakrejda, line 35:
```suggestion
As queries are received by pganalyze, they are parsed, and an internal association is stored that allows looking up query activity by table. Additionally, each query is broken down into [individual scans](/docs/indexing-engine/query-to-scan-breakdown).
```

msakrejda (approved):

schema-statistics/index.mdx
```diff
@@ -0,0 +1,51 @@
...
+backlink_title: 'Documentation'
+---
+
+As part of the "Schema Statistics" functionality, pganalyze allows tracking workload patterns, schema and data model design, as well as changes over time.
+
+## Breakdown of table and index sizes
+
+![](table-treemap.png)
+
+In the schema statistics overview you can find two treemaps, as well as full lists of all the tables and indexes on a given database. The treemap is useful to identify how a specific index or table size relates to the total.
```
keiko713, line 13: Definitely nitpick but I got confused about this "two treemaps", because I only see one treemap _in Tables tab_. I think when you say "overview", you meant all tabs over there, but since there is no tab called "overview", I thought you were talking about the Tables tab.
I think it's okay to leave it as is, but if there is any better way to rewrite this, that'd be great. Uhmm maybe something like "In the schema statistics Tables and Indexes tabs,"?

schema-statistics/index.mdx
keiko713, line 19: Well now I'm getting less confident about leaving this review but I still think having the comma reads better. Ditto to below two paragraphs.

```suggestion
For each table, you can track growth over time, understand how a table utilizes TOAST (The Oversized Attribute Storage Technique) for storing large values, and see the aggregate size for all indexes on the table.
```

schema-statistics/index.mdx
keiko713, line 35: This sentence confuses me a bit. Maybe we can do "As queries are received by pganalyze, they are parsed." then create a next sentence? (that next sentence confuses me a bit though).

keiko713 (approved):
Overall, looks good! Lots of small nitpicks which are totally fine to be disregarded.

schema-statistics/columns-hot-updates.mdx
lfittl: Good point, let me just drop this - people can follow the link and see that its currently in early access.

schema-statistics/columns-hot-updates.mdx
lfittl: Really there are two separate concepts here:
- HOT capable, which refers to the option that a particular update can be a HOT update (HOT capable = there are no indexes on the updated columns)
- HOT update, which is the actual way how the update was performed (requires the update to be HOT capable, and for there to be sufficient space on the page where the existing tuple sits)

In this sentence I'm referring to the second concept, since the lower overhead only applies when its actually a HOT update, not just a HOT capable update. I'll reword from to lowercase "update" though, since I think that's more consistent (updates can be more than just actual UPDATE statements).

schema-statistics/columns-hot-updates.mdx
lfittl: Cool, will reword per your second suggestion. I'll leave out your last suggestion though, since I think we currently always display both statistics datapoints (i.e. top-level table and the inherited stats), even for declaratively partitioned tables.

schema-statistics/columns-hot-updates.mdx
lfittl: The "Status" tab on the Index Advisor, let me see if I can refine the wording.

schema-statistics/index.mdx
lfittl: Eh, I think its probably fine, and not sure how I can reword the next sentence. Will leave as is for now.

PR #217, opened 2024-02-21
Docs for get_relation_stats_ext helper func

keiko713:
Similar to https://pganalyze.com/docs/install/troubleshooting/column_stats_helper, create a dedicated page talking about this helper function. The warning message that mentioned in the doc will be appearing in the pganalyze app in the future.

Unlike the column stats helper, missing this helper function won't cause any warnings in the collector side, due to how we collect these data. Because of that, having this doc under install/troubleshooting is a bit inappropriate, but I can't think of any other place so I think it's okay?

### DIFF
--- a/components/MonitoringUserSetupInstructions.tsx
+++ b/components/MonitoringUserSetupInstructions.tsx
@@ -17,6 +17,7 @@ const MonitoringUserSetupInstructions: React.FunctionComponent<Props> = ({
     <>
       <MonitoringUserBase password={password} noPgMonitor={noPgMonitor} />
       <MonitoringUserColumnStats username="pganalyze" adminUsername={adminUsername} />
+      <MonitoringUserExtStats username="pganalyze" adminUsername={adminUsername} />
     </>
   );
 };
--- a/components/MonitoringUserSetupInstructions.tsx
+++ b/components/MonitoringUserSetupInstructions.tsx
@@ -83,7 +83,7 @@ $$
 $$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}
       </CodeBlock>
       <p>
-        <strong>Note:</strong> We never collect actual table data through this method (from the <code>pg_stats_ext</code> view, we are not collecting things like <code>most_common_vals</code> as you can see in the function definition), but we do collect statistics about the distribution of values in your tables.
+        <strong>Note:</strong> We never collect actual table data through this method (we omit fetching fields like <code>most_common_vals</code> from the <code>pg_stats_ext</code> view, as you can see in the function definition), but we do collect statistics about the distribution of values in your tables.
         You can skip creating the <code>get_relation_stats_ext</code> helper function if the database
         contains highly sensitive information and statistics about it should not be collected.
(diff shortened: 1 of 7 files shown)


components/MonitoringUserSetupInstructions.tsx
```diff
@@ -45,8 +46,47 @@ $$
...
+) AS
+$$
+  /* pganalyze-collector */ SELECT statistics_schemaname, statistics_name,
+  (row_to_json(se.*)::jsonb ->> 'inherited')::boolean AS inherited, n_distinct, dependencies,
+  most_common_val_nulls, most_common_freqs, most_common_base_freqs
+  FROM pg_catalog.pg_stats_ext se;
+$$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}
+      </CodeBlock>
+      <p>
+        <strong>Note:</strong> We never collect actual table data through this method (from the <code>pg_stats_ext</code> view, we are not collecting things like <code>most_common_vals</code> as you can see in the function definition), but we do collect statistics about the distribution of values in your tables.
```
msakrejda, line 86:
```suggestion
        <strong>Note:</strong> We never collect actual table data through this method (we omit fetching fields like <code>most_common_vals</code> from the <code>pg_stats_ext</code> view, as you can see in the function definition), but we do collect statistics about the distribution of values in your tables.
```

index-advisor/getting-started.mdx
```diff
@@ -65,6 +65,20 @@ for warnings in the collector test output (`pganalyze-collector --test
 --verbose`).
 
 
+## Set up the extended stats helper for better recommendations
+
+Extended statistics are used by pganalyze to improve Index Advisor
```
msakrejda, line 70: This might be a good place to link to https://www.postgresql.org/docs/current/sql-createstatistics.html . We could drop the parenthetical "created via..." below if we do that.

index-advisor/getting-started.mdx
```diff
@@ -65,6 +65,20 @@ for warnings in the collector test output (`pganalyze-collector --test
...
+recommendations. If you are using the extended statistics (created via
+`CREATE STATISTICS`), for best results, make sure you are running collector
+0.53.0 or later, and [set up the get_relation_stats_ext helper
+function](/docs/install/troubleshooting/ext_stats_helper).
+
+You can check which collector version you are running on the Settings page for
+your server, and can check if the column stats helpers are installed by looking
```
msakrejda, line 77: This talks about the column stats helper. Also, you mentioned in the PR description that a missing ext stats helper won't cause warnings, so one of those is probably wrong?

msakrejda (approved):
Looks good, apart from one reference to column stats helpers when it should be ext stats helpers. I think putting this in troubleshooting in spite of no warnings is okay, but if we don't have a warning, I don't think we should say things like "check if the [ext stats] helpers are installed by looking for warnings in the collector test output."

index-advisor/getting-started.mdx
keiko713: Good catch, will drop the second half of the sentence.

PR #219, opened 2024-02-23
Update screenshots of AWS management console

keiko713:
There are many more outdated screenshots (not only AWS), but these three images are particularly bothering me (esp. that Postgres9.5 part) so let's update to the latest management console UI ones.

* Replaces 3 screenshots
* Delete one screenshot about making IAM role
   * For this one, AWS updated the UI and there is no longer such "creation summary" page. Instead, they made the creation step for the several steps, so I added what exactly to select in each steps and that should be clear enough


explain/setup/azure_database/03_review_settings.mdx
```diff
@@ -7,8 +7,6 @@ backlink_title: 'Automated EXPLAIN: Setup'
 
 import PgSettingsRecommendations, { getAutoExplainSettingRecommendations } from "../../../components/PgSettingsRecommendations";
 
-import imgRdsParamGroup2 from '../../../images/rds_parameter_group2.gif'
-
```
keiko713, line 11: This was wrongly imported (likely from the copy&paste), it's not used so I removed it.

install/amazon_rds/03_setup_iam_policy.mdx
```diff
@@ -73,7 +69,13 @@ We recommend naming the policy "pganalyze" or similar, so you can easily identif
...
 for pganalyze, and assign the policy to this new role.
 
-<ImgRdsSetUpIamPolicy />
+For the trusted entity, pick the following:
+
+- Trusted entity type: AWS service
+- Use case: EC2
+
+For the permission, choose the policy you just created above.
+We recommend naming the role also "pganalyze" or similar.
```
msakrejda, line 78:
```suggestion
We recommend also naming the role "pganalyze" or similar.
```

msakrejda (approved):

PR #222, opened 2024-03-06
Clarify EC2 instance size requirements

seanlinsley:
The differences are that `t3.micro` has 1 GB of RAM instead of 2, and costs $7.59 a month instead of $15.18.

### DIFF
--- a/install/amazon_rds/04_install_collector_ec2.mdx
+++ b/install/amazon_rds/04_install_collector_ec2.mdx
@@ -17,11 +17,13 @@ For this step, you first need to choose whether you want to install the collecto
 
 The [pganalyze collector](https://github.com/pganalyze/collector) is a daemon process that continuously collects database statistics, and submits them to the pganalyze dashboard in recurring intervals.
 
-You can either run it on a small instance (a `t3.small` suffices), or add it to an existing EC2 instance in your environment.
+You can either run it on a small instance, or add it to an existing EC2 instance in your environment.
+
+When monitoring a single medium-sized database `t3.micro` is usually large enough, though you may need `t3.small` to monitor multiple large databases.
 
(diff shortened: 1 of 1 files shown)


msakrejda (commented):
This reverts https://github.com/pganalyze/pganalyze/pull/982 -- is that worth doing? Collector OOMs are not that easy to diagnose, and may include support interactions, creating an onboarding hurdle. It's a difference of $7/month with on-demand pricing.

Maybe instead we should add a separate docs page on sizing your collector instance, and go into more detail there (and link from here)?

seanlinsley:
Since https://github.com/pganalyze/collector/pull/384/commits/65d04dbfec138a9e61b29af0fd69e840336d6f0a our production collector (which monitors two servers) has reliably stayed below 1 GB. In the image below it peaks at 545 MB.

I'm not sure what benefit a dedicated docs page would bring. Is there really that much to cover?

Screenshot 2024-03-06 at 11 56 39 AM

lfittl:
> Since [pganalyze/collector#384](https://github.com/pganalyze/collector/pull/384) our production collector (which monitors two servers) has reliably stayed below 1 GB. In the image below it peaks at 545 MB.
> 
> I'm not sure what benefit a dedicated docs page would bring. Is there really that much to cover?

I'd also be in favor making this a separate page, and continuing to default to the larger instance in the docs.

The cost difference here is negligible for most organizations, and we should have a place anyway where we can clarify that the collector needs more resources (both on the host and the systemd unit file) if you monitor many servers. If someone is truly cost conscious on this level, they could go to that page and decide to go smaller too.

lfittl:
Closing this for now - we may come back to this but in any case it'd need a rebase over recent docs changes.

PR #223, opened 2024-03-12
Add Query Performance section

keiko713:
This PR adds a new docs section about Query Performance, also will add a dedicated page to explain about query tags.


query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+backlink_href: /docs/query-performance
+backlink_title: 'Query Performance'
+---
+
+Query tags are key-value pairs of runtime information of a query that you can
+add to the SQL statements as a query comment. The runtime information can be a
+name of the background job that is running the query, or a trace ID.
+
+Query tags help to find out where the query is coming from, or also help to find
+the corresponding query of the specific background job or a slow query.
```
msakrejda, line 12: I think we could also add another example here:
```suggestion
Query tags can help determine where a query is coming from, support [tracing](https://pganalyze.com/docs/opentelemetry), or help track down the specific background job corresponding to a slow query.
```
Also, I don't quite follow the background job example, so I reworded that slightly.

msakrejda: Ah, I see, you already link that below. I think it's still okay to use as an example, but having another example would be fine, too.

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+augment SQL statements before execution, with comments containing information
+about the code that caused its execution. Rails 7+ has a built-in feature called
+[Query Logs](https://api.rubyonrails.org/classes/ActiveRecord/QueryLogs.html)
+that allows you to add tags like `controller` and `action` to the query so that
+you can identify which controller and action the query was called from.
+
+In pganalyze, you can see tags of a query in the "Query Tags" tab in the
+individual query details page.
+
+This query tag is also used in the OpenTelemetry integration to specify the
```
msakrejda, line 25:
```suggestion
Query tags are also used in the OpenTelemetry integration to specify the
```

query-performance/query-tags.mdx
msakrejda: Oh actually you're linking external docs below. It might be nice to link to our docs on OpenTelemetry somewhere (either an example here or somewhere else in this page), since otel is a big reason to use query tags.

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+* Query samples
+  * When Log Insights is enabled and integrated, for queries with the execution
+    time exceeding either the following condition will appear in the logs and
+    query tags of these queries will be parsed
+    * `log_min_duration_statement`
+    * `auto_explain.log_min_duration`, if using `auto_explain`
+* Queries appearing in the Connections page
+  * The pganalyze collector collects the currently running queries using
+    `pg_stat_activity`, every 10 seconds
+  * These queries will be shown in the Connection page, also query tags of them
```
msakrejda, line 61:
```suggestion
  * These queries will be shown in the Connections page and their query tags will be parsed
```

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+well as in the "Query Tags" tab.
+
+![Query Tags tab](query_tags.png)
+
+Query tags are parsed from queries either of the following places:
+
+* Query samples
+  * When Log Insights is enabled and integrated, for queries with the execution
+    time exceeding either the following condition will appear in the logs and
+    query tags of these queries will be parsed
```
msakrejda, line 55: Do we need "enabled" for Log Insights? You can disable log collection, but it seems pretty obvious this won't work then, and it's arguably temporarily "not integrated" at that point.

```suggestion
  * When Log Insights is integrated, queries with execution
    time exceeding either of the following settings will appear in the logs and
    their query tags will be parsed
```

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+    * `log_min_duration_statement`
+    * `auto_explain.log_min_duration`, if using `auto_explain`
+* Queries appearing in the Connections page
+  * The pganalyze collector collects the currently running queries using
+    `pg_stat_activity`, every 10 seconds
+  * These queries will be shown in the Connection page, also query tags of them
+    will be parsed
+
+Even if all queries contain the query tags using application-side libraries, it
```
msakrejda, line 64:
```suggestion
Even if all queries contain query tags, it
```
The source doesn't really matter, right? If you're hand-writing SQL, it's the same thing.

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+is possible that it's not shown in the query detail page if there is no query
```
msakrejda, line 65:
```suggestion
is possible that it's not shown in the query detail page if there are no query
```

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+samples for that query, or that query is not running frequent or long enough to
+be appearing in the Connections page.
```
msakrejda, line 67:
```suggestion
samples for that query, or that query is not running frequently enough or long enough to
appear in the Connections page.
```

query-performance/query-tags.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+All the query matching the above condition are counted as a "sample", and you
+can see the details, such as the ratio of values of each tag key, in the Query
+Tags tab.
```
msakrejda, line 71:
```suggestion
All queries matching the above condition are counted as "samples", and you
can see their details, including the frequency of different values for each tag key, in the Query
Tags tab.
```

msakrejda (approved):
Some wording suggestions, but looks great! Thanks for working on this.

query-performance/index.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+* query1: total time 8,000ms for 10 calls, averaging 800ms
+* query2: total time 8,000ms for 20 calls, averaging 400ms
+* query3: total time 9,000ms for 30 calls, averaging 300ms
+* query4: total time 8,000ms for 40 calls, averaging 200ms
+
+This process generates a dataset like `[200ms (40 times), 300ms (30 times), 400ms (20 times), 800ms (10 times)]`.
+he median is 300ms, but this method might not fully represent outliers. For
+instance, if query1 had two exceptionally slow runs (2,000ms each), the true
+99th percentile might be 2,000ms, not 800ms as our averaged data would suggest.
```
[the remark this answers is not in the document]
keiko713: Just realized that the line 39 is missing "T" ("he median is" v.s. "The median is"), thanks for the compliment so that I was able to realize it :P

query-performance/index.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+For example, with query statistics over 10 minutes:
+
+* query1: total time 8,000ms for 10 calls, averaging 800ms
+* query2: total time 8,000ms for 20 calls, averaging 400ms
+* query3: total time 9,000ms for 30 calls, averaging 300ms
+* query4: total time 8,000ms for 40 calls, averaging 200ms
+
+This process generates a dataset like `[200ms (40 times), 300ms (30 times), 400ms (20 times), 800ms (10 times)]`.
+he median is 300ms, but this method might not fully represent outliers. For
```
keiko713, line 39:
```suggestion
The median is 300ms, but this method might not fully represent outliers. For
```

PR #224, opened 2024-03-15
Update the list of collector packages

keiko713:
Just realized that this list is super outdated, so I wanted to update. Since there are so many options with yumyum now, the list is a bit too crowded but I think it's okay.

image

The list is coming from here:

https://github.com/pganalyze/collector/blob/main/packages/repo/sync_deb.sh
https://github.com/pganalyze/collector/blob/main/packages/repo/sync_rpm.sh
https://github.com/pganalyze/collector/blob/main/packages/test/Makefile

### DIFF
--- a/components/CollectorPkgInstallInstructions.tsx
+++ b/components/CollectorPkgInstallInstructions.tsx
@@ -38,12 +38,12 @@ const CollectorPkgInstallInstructions = () => {
 
 type YumProps = {
   kind: 'yum'
-  distro: 'el/8' | 'el/7' | 'fedora/35'
+  distro: 'el/9' | 'el/8' | 'el/7' | 'fedora/37' | 'fedora/36'
 }
 
 type DebProps = {
   kind: 'deb'
-  distro: "ubuntu/jammy" | "ubuntu/focal" | "ubuntu/bionic" | "debian/buster" | "debian/stretch"
+  distro: "ubuntu/jammy" | "ubuntu/focal" | "debian/bookworm" | "debian/bullseye"
 }
 
 type Props = YumProps | DebProps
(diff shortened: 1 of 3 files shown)


components/CollectorPkgInstallInstructions.tsx
```diff
@@ -54,19 +54,21 @@ const CollectorDistroInstallInstructions: React.FunctionComponent<Pick<Props, 'k
   switch (kind) {
     case 'yum':
       installOpts = [
+        [ 'el9', 'el/9', 'RHEL / Rocky / OL 9' ],
```
keiko713, line 57: Interestingly, this is the only place we mention "Oracle Linux". Is it really necessary? I don't think we even test with it really (technically we don't test with Rocky with EL9 too).

components/CollectorPkgInstallInstructions.tsx
```diff
@@ -54,19 +54,21 @@ const CollectorDistroInstallInstructions: React.FunctionComponent<Pick<Props, 'k
...
         [ 'el8', 'el/8', 'RHEL / Rocky / OL 8' ],
         [ 'el7', 'el/7', 'RHEL / CentOS 7' ],
+        [ 'al2023', 'el/9', 'Amazon Linux 2023' ],
```
keiko713, line 60: With our install script, we pick el/9 for amazon linux 2023: https://github.com/pganalyze/collector/blob/94bef78c0dc84b049d963c22692094482d085269/contrib/install.sh#L62-L65

components/CollectorPkgInstallInstructions.tsx
msakrejda: We did add it intentionally: if you're running OL9 I guess you may not know it's a RHEL derivative? I agree it makes this busier, but that clarity might be worth it.

msakrejda (approved):

PR #225, opened 2024-03-15
Update package OS support docs to match collector install script

msakrejda:
See https://github.com/pganalyze/collector/blob/main/contrib/install.sh#L54-L130


msakrejda:
Great minds think alike, but @keiko713's mind is a little faster. Closing in favor of #224.

PR #226, opened 2024-03-19
Update getting started overview page

keiko713:
This PR does 3 things:

* Update getting started overview: update the "what the collector is" to match to the docs overview page
* ~Update Enabling pg_stat_statements page: this contained some old information, so updated to match modern system~
* Fix typos: "where your Postgres is running" to "where your collector is running"

### DIFF
--- a/_install_public.mdx
+++ b/_install_public.mdx
@@ -1,7 +1,7 @@
 In order to use pganalyze you need to install the [pganalyze collector](https://github.com/pganalyze/collector)
 in your environment, either on the database server directly, or a container / VM
-close to it when using managed database services. The collector sends query data
-and other statistics on a continuous basis to the pganalyze cloud service, or
-to your pganalyze Enterprise Server installation.
+that can connect to it when using managed database services. The collector sends
+query data and other statistics on a continuous basis to the pganalyze cloud
+service, or to your pganalyze Enterprise Server installation.
 
 The minimum supported PostgreSQL version is 10.
--- a/_install_public.mdx
+++ b/_install_public.mdx
@@ -1,7 +1,10 @@
 In order to use our service you need to install the [pganalyze collector](https://github.com/pganalyze/collector) on
 your Postgres database server.
 
-The collector sends normalized query data and other statistics to our service
-every 10 minutes.
+In order to use pganalyze you need to install the [pganalyze collector](https://github.com/pganalyze/collector)
+in your environment, either on the database server directly, or a container / VM
+close to it when using managed database services. The collector sends query data
+and other statistics on a continuous basis to the pganalyze cloud service, or
+to your pganalyze Enterprise Server installation.
 
 The minimum supported PostgreSQL version is 10.
--- a/_install_public.mdx
+++ b/_install_public.mdx
@@ -1,6 +1,3 @@
-In order to use our service you need to install the [pganalyze collector](https://github.com/pganalyze/collector) on
-your Postgres database server.
-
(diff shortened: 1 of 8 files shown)


install/01_enabling_pg_stat_statements.mdx
```diff
@@ -21,12 +21,10 @@ On your database server, make sure that the extensions package is installed.
 
 Debian/Ubuntu:
 
-```
-# Replace 9.X with your installed Postgres version:
-sudo apt-get install postgresql-contrib-9.X
-```
+Debian/Ubuntu includes modules like `pg_stat_statements` in the main package,
+therefore no additional install is required.
```
keiko713, line 25: I double checked this but as Maciek mentions below places, this is no longer needed.

https://github.com/pganalyze/pganalyze-docs/pull/124
https://postgrespro.com/list/thread-id/2517774

install/01_enabling_pg_stat_statements.mdx
```diff
@@ -21,12 +21,10 @@ On your database server, make sure that the extensions package is installed.
...
 
-RedHat/CentOS/SLES:
+RedHat/CentOS:
```
keiko713, line 27: I confirmed that this is still required for RHEL. I removed SLES there since we don't mention anywhere else (also I don't think SLES uses yum, looks like it uses zypper).

_install_public.mdx
```diff
@@ -1,7 +1,7 @@
-In order to use our service you need to install the [pganalyze collector](https://github.com/pganalyze/collector) on
-your Postgres database server.
-
-The collector sends normalized query data and other statistics to our service
-every 10 minutes.
+In order to use pganalyze you need to install the [pganalyze collector](https://github.com/pganalyze/collector)
+in your environment, either on the database server directly, or a container / VM
+close to it when using managed database services. The collector sends query data
```
msakrejda, line 3:
```suggestion
in your environment, either on the database server directly, or a container / VM
that can connect to it when using managed database services. The collector sends query data
```
maybe? I think "close to it" is vague (I'm guessing "close" in terms of network topology, but it's not clear), and not strictly required: if you really want your collector running in a different region, that's not ideal, but it should work.

msakrejda: We could also consider
```suggestion
in your environment, either on the database server directly, or a container / VM
that can connect to it. The collector sends query data
```
since you may want to or need to monitor your self-managed database server with the collector installed elsewhere. That's probably a niche configuration, though, so we could keep the "managed services" part.

msakrejda (approved):
Looks good. @lfittl there's no concern about the copy changes to that SEO-juicy pg_stat_statements page, right?

lfittl:
> Looks good. @lfittl there's no concern about the copy changes to that SEO-juicy pg_stat_statements page, right?

Thanks for flagging this. It might be best if we can split out that change from the other ones, since https://pganalyze.com/docs/install/01_enabling_pg_stat_statements is the page that currently ranks pretty well.

Since we're removing content from the page (which could make it too short to be considered relevant), it might be best if we add some additional content that is helpful for potential readers. For example, common issues that one has with pg_stat_statements / what else to know.

_install_public.mdx
keiko713: I'd say that we can keep the managed database services part. I understand your point and it is true, but saying "managed database services" there covers the majority cases and they can feel like "yes that's my case", which is helpful.

RE: changing "close to" to "that can connect to", I was actually thinking the same thing too, so I'm +1 👍 This is copied from the Overview overview page so I'll update there too

PR #227, opened 2024-03-19
Update RDS step 1 (enabling pg_stat_statements)

keiko713:
This mainly update two pages, to match closer for in-app setup steps.
You can find the modified pages in here:

https://pganalyze.com/docs/install/amazon_rds/01_configure_rds_instance

image

https://pganalyze.com/docs/install/troubleshooting/rds_pg_stat_statements_shared_preload_libraries

image

### DIFF
--- a/install/amazon_rds/_01_public.mdx
+++ b/install/amazon_rds/_01_public.mdx
@@ -30,7 +30,7 @@ but the following is the example of what to modify.
  Name | New Value | &nbsp;
  ---- | ------ | -------
  [pg\_stat\_statements.track](https://www.postgresql.org/docs/current/pgstatstatements.html#PGSTATSTATEMENTS-CONFIG-PARAMS) | ALL | Optional, enables tracking of queries inside stored procedures
- [track\_io\_timing](https://www.postgresql.org/docs/current/runtime-config-statistics.htmlR) | 1 | Optional, enables tracking of per-query I/O statistics
+ [track\_io\_timing](https://www.postgresql.org/docs/current/runtime-config-statistics.html) | 1 | Optional, enables tracking of per-query I/O statistics
  [track\_activity\_query\_size](https://www.postgresql.org/docs/current/runtime-config-statistics.html) | 2048 | Optional, increases the max size of the query strings Postgres records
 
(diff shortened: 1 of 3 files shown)


install/amazon_rds/_01_public.mdx
```diff
@@ -16,25 +8,35 @@ This guide assumes you have an already running Amazon RDS PostgreSQL server you
...
 
  Name | New Value | &nbsp;
  ---- | ------ | -------
- [pg\_stat\_statements.track](https://www.postgresql.org/docs/current/pgstatstatements.html#id-1.11.7.39.9) | ALL | Optional, enables tracking of queries inside stored procedures
-[shared\_preload\_libraries](https://www.postgresql.org/docs/current/runtime-config-client.html#RUNTIME-CONFIG-CLIENT-PRELOAD) | pg\_stat\_statements
- [track\_io\_timing](https://www.postgresql.org/docs/current/runtime-config-statistics.html#RUNTIME-CONFIG-STATISTICS-COLLECTOR) | 1 | Optional, enables tracking of per-query I/O statistics
-
----
+ [pg\_stat\_statements.track](https://www.postgresql.org/docs/current/pgstatstatements.html#PGSTATSTATEMENTS-CONFIG-PARAMS) | ALL | Optional, enables tracking of queries inside stored procedures
+ [track\_io\_timing](https://www.postgresql.org/docs/current/runtime-config-statistics.htmlR) | 1 | Optional, enables tracking of per-query I/O statistics
```
msakrejda, line 33:
```suggestion
 [track\_io\_timing](https://www.postgresql.org/docs/current/runtime-config-statistics.html) | 1 | Optional, enables tracking of per-query I/O statistics
```

msakrejda (approved):
Looks good, but I don't recall why we have such a big discrepancy here in the first place...

lfittl:
> Looks good, but I don't recall why we have such a big discrepancy here in the first place...

I suspect there is historic context here that this page ranked highly on Google and we didn't want to break it unnecessary. That said for this page (which lies in the common path that RDS users go through), I think we should prioritize the user experience over any SEO benefit.

PR #228, opened 2024-03-19
Guided setup: update supported Linux distributions

msakrejda:


msakrejda:
Actually, it looks like this was also fixed in #224. I guess we just hadn't deployed that when we went through the install this morning.

PR #232, opened 2024-03-20
Fix extended statistics monitoring helpers for Postgres 12

msakrejda:
Our get_relation_stats_ext helper is specified as returning a row that
starts with two text fields, but these come from the pg_stats_ext view
and are actually `name` columns (the type for Postgres identifiers).

Postgres 13 introduced an implicit name-to-text cast, so in that version
and newer versions, our helper function works fine, but on Postgres 12
it fails with

ERROR:  return type mismatch in function declared to return record
DETAIL:  Final statement returns name instead of text at column 1.
CONTEXT:  SQL function "get_relation_stats_ext"

Update the function definition to use explicit casts for the first two
columns.


msakrejda:
See also https://github.com/pganalyze/collector/pull/517

lfittl (approved):

PR #233, opened 2024-03-22
Update install choice buttons

keiko713:
This PR does the redesign of the picking up provider buttons (install choice buttons). Currently, design is pretty different between 1) docs Overview page 2) Get Started page 3) In app. With this, we unify the design so all the places look pretty the same.

1) docs Overview page (width: more than 980px, 3 columns)
image

2) Get Started page (width: between 620px and 980px, 2 columns)
image

3) In app (width: more than 980px, 3 columns)
image

4) width less than 620px, single column
image


lfittl:
> This PR does the redesign of the picking up provider buttons (install choice buttons). Currently, design is pretty different between 1) docs Overview page 2) Get Started page 3) In app. With this, we unify the design so all the places look pretty the same.

Thanks for updating this!

The one thing that sticks out for me is the extra white space around most of the buttons (i.e. those without two lines of description) - it doesn't feels right visually to me.

I wonder if we should require that the description texts are one line maximum after all?

jensy:
> The one thing that sticks out for me is the extra white space around most of the buttons (i.e. those without two lines of description) - it doesn't feels right visually to me.

That is intentional, to not have to worry about the length of the description text. In my design I vertically centered the description text tho, and I think that helps visually.

Screenshot for reference

image

If you feel strongly about, I am happy to forfeit this tho, but I am sure this will bite us later, and I rather go for clarity in communication than optimizing for visual things.

msakrejda:
It would be nice to have the cards be uniform height, like in the design (whether through setting a min-height or limiting to a single line of text; I don't feel strongly about that).

Also, I'm not sure if we need the "and many other environments" on self-managed if we are explicitly going to add a note like "Don't see your platform listed? Learn how to integrate Postgres on other providers." below.

keiko713:
Re: vertical alignment, I was lazy about it so good job everyone for pointing that out. Here is some single line version (by making it shorter for both Azure and self-managed per Lukas/Maciek's suggestion):

image

It's simple, but this could still be two lines description when the width got smaller, and the button size can be different depending on two lines v.s. one line:

image

Here is a min-height version, with min-height being 38px + line-height 19px:

image

image

Which one do people prefer? 1) line-height 100% 2) min-height 38px?
Also should we cut the description part to be a single line with the normal screen size?

msakrejda:
Nice, this looks awesome. I prefer the min-height version, but +1 to either one (the other cleanup makes the height difference stand out less, I think). It might also be worth testing with different levels of browser zoom to make sure nothing breaks in the desired approach (that should trump any aesthetic concerns).

Not sure what you mean by cutting the description and which is the normal screen size. If it's whether to cut it to what you have in the screenshots, +1: that works great. If it's too cut it further for the second, narrower screen size: I don't think that's necessary.

keiko713:
> Nice, this looks awesome. I prefer the min-height version, but +1 to either one

I'm currently adopting the min-height version, I think this works well too 👍 

> It might also be worth testing with different levels of browser zoom to make sure nothing breaks in the desired approach

I checked with Chrome, Safari, Firefox and all of them act pretty much the same. When to change from buttons count in one row (there are 3, 2, 1 cases) is still a bit questionable, but I don't think it's too off. You can checkout the review app for this too:

https://pganalyze-we-install-ch-eglz0u.herokuapp.com/docs

> Not sure what you mean by cutting the description and which is the normal screen size

Sorry, I completely understand your confusion 🤦 (I don't even know what I meant).
I _think_ I was asking if adopting Lukas/Maciek's feedback for the shorter version (e.g. dropping "for PostgreSQL" for Cosmos) look okay to everyone.

style.module.scss
```diff
@@ -36,45 +91,32 @@
...
-
 .overviewFeatureChoiceStep {
   flex-direction: column;
 }
 
-@media (min-width: 620px) {
-  .overviewInstallChoiceStep img {
-    order: 0;
-  }
+@media (min-width: 640px) {
```
keiko713, line 98: [note for reviewers] I'm changing the min-width of this. I'm not sure where 620px came from, but I think standard is more like 640px (`sm`). I'm picking this up from Tailwind: https://tailwindcss.com/docs/responsive-design

I also used 768px as min-width to become 3 buttons, which is `md` in Tailwind, as well as the starting size of tablets and small desktops.

style.module.scss
```diff
@@ -36,45 +91,32 @@
...
+    grid-template-columns: 1fr 1fr;
   }
   .overviewFeatureChoice {
     grid-template-columns: 1fr 1fr;
   }
   .overviewInstallChoiceStep, .overviewFeatureChoiceStep {
     flex-direction: column;
   }
+  .overviewInstallChoiceProviderText {
+    min-height: 38px;
```
keiko713, line 112: [note for reviewers] I only have min-height here, which will be applied to 2 or 3 buttons in a row case. For the 1 button in a row case, I found that 38px height here was too much space (need to scroll a lot in mobile, for example). 19px height looks good to me for 1 button in a row case, so I tweaked here.

keiko713:
Ok updated based on the PR review sessions today. The major thing is to align install and feature choice aligned, in the Overview page. Attached screenshots for both large screen + mobile, but you can also check out the deploy preview in https://pganalyze-we-install-ch-eglz0u.herokuapp.com/docs/

image

image

keiko713:
And this is how it looks in-app:

image

style.module.scss
lfittl: I'm not sure if this is the right change I'm commenting on, but if it is: I wonder if this wouldn't be better without the `min-height`, to reduce the whitespace:

Screenshot 2024-04-02 at 9 55 22 PM

lfittl (approved):

style.module.scss
lfittl: Actually, I just realized this is related to the earlier comment by you asking re: line-height vs min-height and I'm late to comment on that - this is not a strong opinion, so feel free to merge as-is (we can revise in a follow-up PR).

keiko713: Yeah let's go with as-is, happy to adjust later 👍
Also Jens haven't really checked the latest version, so he may want to say something about it too, which I can adjust as a follow-up if needed

jensy:
No, no, I kept an eye on it, but did not have any other feedback. Looks great!

PR #234, opened 2024-04-02
Add Step 0: Overview page for the AWS setup steps

keiko713:
This PR adds step 0 for the AWS setup, to explain what is needed prior to setting up.
We plan to add some diagram to this page later on too, to give users better understanding for what they're going to set up.

In addition to adding the step 0, I also combined in-app and public doc of step 1. This one doesn't really need to be different between in-app and public, so I think it'd be better to combine. This results that the in-app version increased the content a bit (re: config params, enhanced monitoring) but I think that's a nice thing to mention there.

I also would love to improve the "Review pg_stat_statements configuration parameters" section. We should use the same component for this section across the all providers, so we could potentially make one page for this (guess what, this can be in https://github.com/pganalyze/pganalyze-docs/pull/230 even). Datadog has nice doc talking about this part, I'd love to have something similar: https://docs.datadoghq.com/database_monitoring/setup_postgres/rds/?tab=postgres15#configure-postgres-settings
In addition to it, we can talk about how to change these settings depending on the provider (e.g. in AWS, they should change the parameter group). Well, that'd be the follow-up PR.


install/amazon_rds/00_overview.mdx
```diff
@@ -0,0 +1,44 @@
...
+install_track_title: Installation Guide (Amazon RDS and Amazon Aurora)
+backlink_href: /docs/install
+backlink_title: 'Installation Guide'
+---
+
+To monitor your Amazon RDS for PostgreSQL or Amazon Aurora PostgreSQL servers
+with pganalyze, you need to install the pganalyze collector on an Amazon EC2
+instance, with Amazon ECS, or with Docker.
+The collector running in your environment sends query data and other statistics
+on a continuous basis to pganalyze.
```
msakrejda, line 12: The current wording is a little awkward, since it could be read like "on an EC2 instance... with ECS or with Docker" which doesn't really make sense. Breaking this out into a separate sentence makes it easier to follow, but my version could probably still be improved.

We should also avoid saying "query data", since users could get the mistaken impression that query results are sent to pganalyze.

```suggestion
To monitor your Amazon RDS for PostgreSQL or Amazon Aurora PostgreSQL servers
with pganalyze, you need to run the pganalyze collector to connect to your database
and send query metadata and other statistics on a continuous basis to pganalyze.
The collector can be installed directly on an Amazon EC2 instance, with Amazon ECS,
or with Docker.
```

install/amazon_rds/00_overview.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+## Prerequisites
+
+**Database Server**
+* Amazon RDS for PostgreSQL or Amazon Aurora PostgreSQL server to monitor
+* PostgreSQL version is 10 and above
```
msakrejda, line 18: The preceding bullet point is not a full sentence, so we can just make this
```suggestion
* PostgreSQL version 10 or above
```
(Also, totally nit-picking, but I think "or above" makes more sense than "and above" in this case: we support versions "10 and above", but the prerequisite is the version of your specific database, which will be just one version, so "10 or above" because there's just one.)

msakrejda (approved):

install/amazon_rds/00_overview.mdx
keiko713: I'm going to merge as your version (to reduce the rebasing pain) but here is some other ideas:

"The collector can be installed directly on an Amazon EC2 instance or with Amazon ECS, depending on your operational environment."

Basically, I'm dropping Docker here. Docker is one of the options, but feels like it can be dropped from this summary (mainly to simplify and avoid the awkwardness that you feel). When we add EKS, maybe we can say something like:

"The collector can be installed directly on an Amazon EC2 instance, with Amazon ECS, or with Amazon EKS, depending on your operational environment."

Regarding the query data part, +1 on your change, though I think this can also be reworded some, like the following:

"To monitor your Amazon RDS for PostgreSQL or Amazon Aurora PostgreSQL servers with pganalyze, you need to run the pganalyze collector. The pganalyze collector connects to your database to capture query metadata and various statistics, then relays this information back to pganalyze on a continuous basis."

I split this to two sentences, also add the description that the collector capture these data too. What do you think?

msakrejda: Yes. +1 on the second suggestion, especially. One of the "cheat codes" for clearer writing is shorter sentences. And your version flows better, since my sentence just assumes you will infer where the collector gathers these stats.

PR #235, opened 2024-04-04
Update pgss page and create table for recommendation values

keiko713:
In https://github.com/pganalyze/pganalyze-docs/pull/234, I mentioned that I would love to have a better section for the recommendation values of pgss. Here is some draft of how it looks like.

I also updated RDS and Azure docs to point to the page for the recommendations. Ideally we would like to have this link for other providers too, but RDS and Azure are currently only the providers that have good section for this, so I decided to only add to them for now.

The table looks a bit sad with "default: x" part, but I personally wanted to see what the default value was, so I added it. I'm okay removing it if people don't think it's needed.

image


install/01_enabling_pg_stat_statements.mdx
```diff
@@ -32,24 +32,67 @@ sudo yum install postgresql-contrib
...
 
-# Track statements generated by stored procedures as well
 pg_stat_statements.track = all
-
-# Increase the maximum number of statements tracked
 pg_stat_statements.max = 10000
-
-# Enable tracking of per-query I/O statistics
-track_io_timing = 1
+track_io_timing = on
```
msakrejda, line 47: :tada: this way it's clear this is boolean

install/01_enabling_pg_stat_statements.mdx
```diff
@@ -32,24 +32,67 @@ sudo yum install postgresql-contrib
...
+      <td>
+        <a target="_blank" href="https://www.postgresql.org/docs/current/pgstatstatements.html#PGSTATSTATEMENTS-CONFIG-PARAMS">
+          <code>pg_stat_statements.max</code>
+        </a>
+      </td>
+      <td>10000 (default: 5000)</td>
+      <td>
+        Increases the maximum number of statements tracked.
+        Recommended to increase further more if the database tends to have lots of distinct statements.
+        Restart is required for changes to take effect.
```
msakrejda, line 78:
```suggestion
        Increases the maximum number of statements tracked.
        Recommended to increase further if the database tends to have lots of distinct queries.
        Restart is required for changes to take effect.
```
I think "queries" instead of "statements" in one of these sentences helps tie this back to query activity, which is primarily the statements most users will be interested in.

I'm also tempted to add something like
> Queries that have not been seen recently will be pruned once this number is hit. The pruning happens in batches, so you may see the count less than the max even if the max has been reached

because we do get questions about that occasionally. That's probably too much text for the box. Maybe a footnote? Or we can leave it off if you think it doesn't fit anywhere.

install/01_enabling_pg_stat_statements.mdx
```diff
@@ -32,24 +32,67 @@ sudo yum install postgresql-contrib
...
+      <td>on (default: off)</td>
+      <td>Enables tracking of per-query I/O statistics. Might be turned on by default for some managed databases.</td>
+    </tr>
+  </tbody>
+</table>
+
 ## Restart the PostgreSQL daemon
 
-If you change `shared_preload_libraries` above, a restart is required:
+If you change `shared_preload_libraries` or `pg_stat_statements.max` above, a restart is required:
```
msakrejda, line 95: Good catch.

install/azure_database/01_configure_azure_instance.mdx
```diff
@@ -11,25 +11,29 @@ export const ImgAzureServerParams = () => <img src={imgAzureServerParams} />
 
 
 In order to set up pganalyze monitoring for [Azure Database for PostgreSQL](https://azure.microsoft.com/en-us/products/postgresql/) you'll need
-to first follow these steps to enable the [pg\_stat\_statements](http://www.postgresql.org/docs/current/static/pgstatstatements.html)
+to first follow these steps to enable the [pg_stat_statements](http://www.postgresql.org/docs/current/static/pgstatstatements.html)
```
msakrejda, line 14: ?

msakrejda (approved):
Looks good. The default column is a little squished, but I agree that it's useful, so it's fine to leave it.

Does removing the underscore escapes for `pg_stat_statements` not break things? Does it still work both in the app and in public docs?

install/azure_database/01_configure_azure_instance.mdx
keiko713: I believe this question mark is about what you are talking in the overall review comment, if it's okay to remove the escape (underscore). I checked this both in public and in app, both look okay without having the escapes.
Here is one in-app example that we already use "pg_stat_statements" without escaping (nor in the code block), but doing fine: https://staging.pganalyze.com/organizations/pganalyze/servers/install/aiven/01_create_monitoring_user

(I also mentioned this in https://github.com/pganalyze/pganalyze-docs/pull/230#discussion_r1548859435, hehe)

install/01_enabling_pg_stat_statements.mdx
keiko713: I think that "Queries that have not been seen recently will be pruned..." sentence makes sense to have it here. That sentence actually goes nicely with `pg_stat_statements_reset()` and `pg_stat_statements_info` view, so let me try to add some section to talk about it

install/01_enabling_pg_stat_statements.mdx
```diff
@@ -32,24 +32,81 @@ sudo yum install postgresql-contrib
...
+          <code>track_io_timing</code>
+        </a>
+      </td>
+      <td>on (default: off)</td>
+      <td>Enables tracking of per-query I/O statistics. Might be turned on by default for some managed databases.</td>
+    </tr>
+  </tbody>
+</table>
+
+### The `pg_stat_statements_info` view
```
keiko713, line 93: @msakrejda this is a new section added, would love to get your 👀

install/azure_database/01_configure_azure_instance.mdx
msakrejda: Huh, okay. Good to know.

msakrejda (approved):
Updates look good.

PR #236, opened 2024-04-05
Unify install choice button

keiko713:
We introduced a new component called `InstallChoice` in https://github.com/pganalyze/pganalyze-docs/pull/233

With this PR, we use this component for everything about choosing something, to unify the design of it. I also updated all the logos to the latest as some of them were out of date.

Here is how it looks:

image

We have a mix of 3 buttons and 4 buttons currently (above screenshot is 4 buttons), but as we plan to add k8s, most of them will become 4 buttons. 

Here is how it looks in-app:

image


images/logo_ubuntu.png
msakrejda, line 1: Hmm, is the square version of the latest Ubuntu logo from an official source? The one you shared before was a rectangle and so harder for us to use here. Some brands get touchy about edits to their logos, even like this. I wonder if it might not be better to stick with the 2010 version for now...

msakrejda (approved):
Looks great. One question on the Ubuntu logo, but I guess worst case we'd change it if they complain.

images/logo_ubuntu.png
keiko713: Good question. I picked this up from the favicon of https://ubuntu.com/
That said, I love playing safer side and pretty much everywhere I searched is still using 2010 version, so I think I'm gonna do the same and use 2010 version

keiko713: Uhmmm actually I'm gonna keep it as is, I just found a youtube channel of Ubuntu and it's also using the favicon style: https://www.youtube.com/@UbuntuOS/featured
Also 2010 version logo is not exactly easily available, at least there is no official one that I was able to find, so I'll stick with favicon version. Definitely happy to change when/if they complain.

PR #237, opened 2024-04-05
Move IAM policy/role creation to after selecting collector platform

keiko713:
With current flow, setting up IAM policy/role comes before the collector platform selection.

image

This PR removes step 3, then add these steps inside of each platform's section after selected. Also, add the "CLI" and "console" option for IAM policy/role creation so that the user can use CLI to do such thing.

It might be better to review this with review app to see how it looks like. When you do so, please go from step 2 of aws section and see if it makes sense:

https://pganalyze-we-new-aws-in-xuapf2.herokuapp.com/docs/install/amazon_rds/02_create_monitoring_user

image


style.module.scss
```diff
@@ -145,6 +145,7 @@
 
 .navTabs {
   border-bottom: 1px solid #ddd;
+  margin-left: 0px !important;
```
keiko713, line 148: I just noticed that currently we have a sad margin here:

image

This is going to help the nav tab to start without having that margin.

install/amazon_rds/_03_create_iam_role.mdx
```diff
@@ -0,0 +1,31 @@
+import TabPanel, {TabTextContent} from '../../components/TabPanel'
+import CreateIAMRoleCLI from './_03_create_iam_role_cli.mdx'
+import CreateIAMRoleConsole from './_03_create_iam_role_console.mdx'
+
+export const IAMRoleCreation = () => {
+  const tabs = [
+    [ 'cli', 'CLI' ],
+    [ 'console', 'console'],
```
keiko713, line 8: I believe aws is calling the dashboard "console", you can find it in this doc too (it has "console", "AWS CLI", "AWS API"):

https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html

I first thought that "console" is not so obvious to talk about the UI dashboard thingy, but I guess it's okay if it's what aws uses?

install/amazon_rds/_03_setup_iam_policy_cli.mdx
```diff
@@ -0,0 +1,12 @@
+The following command creates a new IAM policy named pganalyze:
+
+```bash
+# First, download the policy JSON file
+curl -O https://pganalyze.com/docs/install/amazon_rds/pganalyze-collector-policy.json
```
keiko713, line 5: TODO: this link doesn't work. With ECS, we also have a long JSON, and we say something like:

> Now, save the following ECS task definition to a file, for example pganalyze_task.json:

We can do the same technically here, though I thought this is a bit more user friendly.
Well, if it's harder to try serving this json file under /docs (I wasn't able to find easy way to do so so far, sadly), I guess we should do the same as what ECS does.

msakrejda: I'm not sure how to serve files under docs. My guess is we'd need to write a Gatsby GraphQL query that will run as part of the build, but looking around briefly, I don't see how to accomplish this. You may have better luck.

Alternately, we could do something like [this](https://github.com/gatsbyjs/gatsby/issues/872#issuecomment-298046002) and just copy the files to the /static/ folder we already have post-build. Thoughts?

keiko713: Thanks for the comment. Yeah I think there is some way to accomplish this, but sounds like it's gonna be complicated so I'll go as the same way as ECS (aka show policy json inline) for now.

msakrejda (approved):
I think we'll need to update paths in the install docs component when incorporating this into the in-app docs, but looks good.

PR #239, opened 2024-04-09
Maintain a single generated password across page views

msakrejda:
Currently, a user returning to an earlier install step to check a
generated password would end up seeing a different password, since
these are regenerated on page view.

Instead, cache the generated result while the page is open.


components/WithGeneratedPassword.tsx
```diff
@@ -0,0 +1,32 @@
...
+    // ensure the password is generated during client-side rendering, to avoid
+    // generating a single recommended password for everyone at build time
+    setPassword(generatePassword());
+  }, [])
+  return (
+    <GeneratedPasswordContext.Provider value={password}>
+      {children}
+    </GeneratedPasswordContext.Provider>
+  );
+}
```
msakrejda, line 26: This component comes from the deleted file GeneratePassword.tsx.

install/aiven/01_create_monitoring_user.mdx
```diff
@@ -5,10 +5,26 @@ backlink_href: /docs/install
 backlink_title: 'Installation Guide'
 ---
 
-import Step01CreateMonitoringUserAiven from "./_01_create_monitoring_user.mdx"
+import MonitoringUserSetupInstructions from "../../components/MonitoringUserSetupInstructions";
```
msakrejda, line 8: These wrapper components are no longer necessary with this new approach. Instead, inline the respective `_...` components into the main ones that included them.

keiko713 (approved):
Overall looks great but got some question. Currently, if I open a new tab in the same browser, or if I reload the page with browser reload button, it'll pick up the different/new password. As long as it's in the same tab (same "page"), the password is preserved.
Just wondering if that's what we all agreed on. I feel that it's slightly inconvenient (as a person who opens up lots of tabs), but I think the opposite is also not ideal (that the password persists until the browser is closed?), so I'm not saying that we need to do the opposite, but just wanted to double check that this is the expected behavior.

msakrejda:
Yeah, that's intentional. I don't think we want to try to persist these generated passwords too hard, since we don't want to lead users to unintentionally reuse generated passwords.

And the actual behavior is actually slightly different in the app and in the public docs. In the public docs, we generate the password once on page load, and keep the same password across client-side navigations. If you reload the page or open a new tab, you'll get a new password. In the app, the context preserving the generated password is actually at the level for each platform selection. So if you look at self-managed instructions, and then go look at RDS instructions, you'll get a different password. While looking at a selected platform (e.g., self-managed), the generated password will persist. I think this is okay, since I don't think it's a common use case to want to preserve the same password across two different platforms.

keiko713:
Sounds good, thanks for explaining!

PR #241, opened 2024-04-10
Improve Heroku auto_explain docs

keiko713:
This update the docs structure of Heroku auto_explain. This follows really close to what we do with Crunchy Bridge (aka lots of copy&pasting), as they also load auto_explain in `session_preload_libraries`.
https://pganalyze.com/docs/explain/setup/crunchy_bridge/01_auto_explain_check

I'll make inline comments of the points that I changed specific to Heroku and want attentions during the review.

Here is how it looks in-app:

https://github.com/pganalyze/pganalyze-docs/assets/911433/68746d99-8a65-484f-815a-f633e29933f6


explain/setup/crunchy_bridge/01_auto_explain_check.mdx
```diff
@@ -5,17 +5,17 @@ backlink_href: /docs/explain/setup
 backlink_title: 'Automated EXPLAIN: Setup'
 ---
 
-import Step01InApp from './_01_in_app.mdx'
-import Step01Public from './_01_public.mdx'
+import Step01InAppCB from './_01_in_app.mdx'
+import Step01PublicCB from './_01_public.mdx'
```
keiko713, line 9: [note for reviewer] Looks like we have problems when we use the same name (e.g. `Step01InApp`) with different contents. I can see that we're adding this pattern (platform specific name at the end) for other places, so I'm doing it here too. Otherwise, at least locally, I was seeing wrongly cached contents.

explain/setup/heroku/03_review_settings.mdx
```diff
@@ -0,0 +1,42 @@
...
+these <AutoExplainLink version={props.pgMajorVersion}>here</AutoExplainLink>.
+The right configuration will depend on your server and workload,
+but we've found the following is a good starting point:
+
+<PgSettingsRecommendations recommendations={getAutoExplainSettingRecommendations(props.settings)} />
+
+Heroku offers to modify several of these settings, but not all of them. Notably,
+you cannot update the `auto_explain.log_format` setting with Heroku Postgres
+currently. Only JSON format is supported by pganalyze and as text format support
+is limited, you may see the errors on some EXPLAIN plans.
```
keiko713, line 29: This is Heroku specific sentence. Heroku currently doesn't offer modifying the `log_format` unfortunately, so we should be noting this (and this would be one downside of migrating to auto_explain, maybe). This sentence and below sentence is something I'd like additional eyes.

explain/setup/heroku/_01_public.mdx
```diff
@@ -0,0 +1,29 @@
+# Check if auto_explain already enabled
+
+First, to check if `auto_explain` is already enabled on your system, run the
+following query with `heroku pg:psql`:
```
keiko713, line 4: With Crunchy, we say "run the following query as a Postgres superuser:", but I thought it's not obvious, so I changed it to `heroku pg:psql`

explain/setup/heroku/03_review_settings.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+There are a number of `auto_explain` settings that allow you to configure which queries
+will generate EXPLAIN plans. You can read the full Postgres documentation about
+these <AutoExplainLink version={props.pgMajorVersion}>here</AutoExplainLink>.
+The right configuration will depend on your server and workload,
+but we've found the following is a good starting point:
+
+<PgSettingsRecommendations recommendations={getAutoExplainSettingRecommendations(props.settings)} />
+
+Heroku offers to modify several of these settings, but not all of them. Notably,
```
msakrejda, line 26:
```suggestion
Heroku allows users to modify several of these settings, but not all of them. Notably,
```

explain/setup/heroku/03_review_settings.mdx
msakrejda, line 29: I think "only JSON format is supported" may be a little strong. Text format is definitely a second-class citizen, but it mostly works, especially if you're not doing anything too weird in your queries. I've used the phrase "best-effort" to describe this in the past, so maybe

```suggestion
currently. Text-format auto_explain output is supported on a best-effort basis, so you may see errors for some EXPLAIN plans.
```

explain/setup/heroku/03_review_settings.mdx
```diff
@@ -0,0 +1,42 @@
...
+you cannot update the `auto_explain.log_format` setting with Heroku Postgres
+currently. Only JSON format is supported by pganalyze and as text format support
+is limited, you may see the errors on some EXPLAIN plans.
+
+
+You can check out [Heroku documentation](https://devcenter.heroku.com/articles/heroku-postgres-settings#auto-explain)
```
msakrejda, line 32:
```suggestion
You can check out the [Heroku documentation](https://devcenter.heroku.com/articles/heroku-postgres-settings#auto-explain)
```

explain/setup/heroku/_01_public.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+```
+SHOW session_preload_libraries;
+```
+
+If `auto_explain` is already there, it is already enabled.
```
msakrejda, line 10:
```suggestion
If `auto_explain` is already there, it is enabled.
```
It feels weird to repeat "already" here.

msakrejda (approved):

explain/setup/heroku/03_review_settings.mdx
keiko713: Yeah agree with you that "only JSON format is supported" is a little strong (but we're actually using it in a tooltip 🙃). Given that other providers can change this config, I think we are fine leaving the tooltip as is, and update here as you suggested (I like the best-effort, as we are definitely doing the best effort there)

explain/setup/heroku/_01_public.mdx
keiko713: good point, I'll also update the other place too 👍

explain/setup/crunchy_bridge/01_auto_explain_check.mdx
keiko713: Uhmm, good point about documenting this. Actually, I've been having a question of how the people are checking if changes that they make "works".
Currently, I'm testing by loading the update in pganalyze-web repo (private) and run `gastby develop` over there but not everyone can do it 🤔 In CONTRIBUTING.md, we could write something around there too.

keiko713:
This will conflict with https://github.com/pganalyze/pganalyze-docs/pull/237 (in-app side) but I decided to merge this first, then I'll take care of that one.

PR #243, opened 2024-04-12
Update the IAM policy link

keiko713:
I just realized that I missed to update several links in https://github.com/pganalyze/pganalyze-docs/pull/237
Now I understand why we wanted to have that page separately... but it's a bit too late to realize, and I think we can live with it. Hopefully.


lfittl (commented):
I wonder if we should instead have a separate page that clearly lists the IAM policy, but isn't tied to any particular installation method?

(I think its also okay to link to the EC2 setup method, but maybe easy if there is just a simple page that has the policy information and nothing else)

msakrejda (commented):
Big +1. I think that'd be much clearer.

keiko713:
Uhmm good point, let me craft something

keiko713:
Going to close this in favor of https://github.com/pganalyze/pganalyze-docs/pull/246

@lfittl @msakrejda could you please check out the new one? Thanks!

PR #244, opened 2024-04-12
Update how to handle Aurora cluster with Docker

keiko713:
The current description is unclear that you need to run two separate Docker container per one simple cluster (writer and reader). It almost looks like you don't need to, which is confusing.
Let's make this clear.

This part was initially added in https://github.com/pganalyze/pganalyze-docs/pull/110

### DIFF
--- a/install/amazon_rds/04_configure_the_collector_docker.mdx
+++ b/install/amazon_rds/04_configure_the_collector_docker.mdx
@@ -12,7 +12,7 @@ import { SettingEnvVarsDocker } from './_04_setting_env_vars.mdx';
 
 Create your configuration file (e.g. named `pganalyze_collector.env`) with environment variables like this:
 
-<SettingEnvVarsDocker />
+<SettingEnvVarsDocker apiKey={props.apiKey} />
 
 **Note:** The pganalyze collector allows for more [optional settings](https://pganalyze.com/docs/collector/settings) (e.g. AWS access keys, multiple database names)
 
--- a/install/amazon_rds/04_configure_the_collector_docker.mdx
+++ b/install/amazon_rds/04_configure_the_collector_docker.mdx
@@ -44,21 +44,20 @@ Create your configuration file (e.g. named `pganalyze_collector.env`) with envir
 
 ## Handling Amazon Aurora clusters vs instances
 
-In the case of Amazon Aurora, the collector automatically resolves `cluster` endpoints to the underlying writer instance, and `cluster-ro` to the underlying reader instance:
+In the case of Amazon Aurora, the collector automatically resolves `cluster` endpoints to the underlying writer instance.
 
 ```
 DB_HOST=mydbcluster.cluster-123456789012.us-east-1.rds.amazonaws.com
 ```
 
-or
+This will only monitor the writer instance. If you also want to monitor a reader instance, you'll need to run another Docker container.
+Create a new configuration file (e.g. named `pganalyze_collector.replica.env`) and specify the `DB_HOST` like the following:
 
 ```
 DB_HOST=mydbcluster.cluster-ro-123456789012.us-east-1.rds.amazonaws.com
 ```
 
-For writer instances this is supported for any size of cluster, but for reader instances this is only supported in two-node clusters (i.e. single reader instance).
-
-If you have multiple readers you want to monitor you instead need to specify each instance endpoint separately, running one pganalyze collector Docker container for each instance:
+If you have multiple readers you want to monitor, you need to specify each instance endpoint separately, running one pganalyze collector Docker container for each instance:
 
 ```
 DB_HOST=mydbinstance1.123456789012.us-east-1.rds.amazonaws.com
--- a/install/amazon_rds/04_configure_the_collector_docker.mdx
+++ b/install/amazon_rds/04_configure_the_collector_docker.mdx
@@ -6,13 +6,15 @@ backlink_title: 'Installation Guide'
 ---
 
 import PublicLastStepLogInsightsLink from '../_public_last_step_log_insights_link.mdx'
-import SettingEnvVarsDockerAWS from './_04_setting_env_vars.mdx';
+import { SettingEnvVarsDocker } from './_04_setting_env_vars.mdx';
 
 ## Configuring the collector with Docker
 
 Create your configuration file (e.g. named `pganalyze_collector.env`) with environment variables like this:
 
-<SettingEnvVarsDockerAWS />
+<SettingEnvVarsDocker />
+
+**Note:** The pganalyze collector allows for more [optional settings](https://pganalyze.com/docs/collector/settings) (e.g. AWS access keys, multiple database names)
 
 ## Handling Amazon Aurora clusters vs instances
 
--- a/install/amazon_rds/04_configure_the_collector_docker.mdx
+++ b/install/amazon_rds/04_configure_the_collector_docker.mdx
@@ -40,7 +40,7 @@ db_host = mydbcluster.cluster-ro-123456789012.us-east-1.rds.amazonaws.com
 "
 ```
 
-Alternatively, you can also run another Docker container for a reader instance.
+Alternatively, you can run a separate Docker container to monitor a reader instance.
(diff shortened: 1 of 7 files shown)


install/amazon_rds/04_configure_the_collector_docker.mdx
```diff
@@ -44,21 +44,20 @@ Create your configuration file (e.g. named `pganalyze_collector.env`) with envir
...
-In the case of Amazon Aurora, the collector automatically resolves `cluster` endpoints to the underlying writer instance, and `cluster-ro` to the underlying reader instance:
+In the case of Amazon Aurora, the collector automatically resolves `cluster` endpoints to the underlying writer instance.
 
 ```
 DB_HOST=mydbcluster.cluster-123456789012.us-east-1.rds.amazonaws.com
 ```
 
-or
+This will only monitor the writer instance. If you also want to monitor a reader instance, you'll need to run another Docker container.
+Create a new configuration file (e.g. named `pganalyze_collector.replica.env`) and use the `cluster-ro` endpoint as the `DB_HOST`:
```
msakrejda, line 54: Technically, you could use CONFIG_CONTENTS to monitor the whole cluster with one collector. We do mention that in [our docs](https://pganalyze.com/docs/collector/settings) (separately, should it be clearer?). I feel like we should mention it here for customers who would rather just run a single collector. How about something like
```suggestion
This will only monitor the writer instance. If you also want to monitor a reader instance, you'll need to use the [`CONFIG_CONTENTS`](https://pganalyze.com/docs/collector/settings) configuration mechanism and add another section with the `cluster-ro` endpoint as the `db_host`.

Alternately, you can run a separate Docker container to monitor each node. Create a new configuration file (e.g. named `pganalyze_collector.replica.env`) and use the `cluster-ro` endpoint as the `DB_HOST`.
```

install/amazon_rds/04_configure_the_collector_docker.mdx
```diff
@@ -44,21 +44,20 @@ Create your configuration file (e.g. named `pganalyze_collector.env`) with envir
...
 
 ```
 DB_HOST=mydbcluster.cluster-ro-123456789012.us-east-1.rds.amazonaws.com
 ```
 
-For writer instances this is supported for any size of cluster, but for reader instances this is only supported in two-node clusters (i.e. single reader instance).
-
-If you have multiple readers you want to monitor you instead need to specify each instance endpoint separately, running one pganalyze collector Docker container for each instance:
+If you have multiple readers you want to monitor, you need to specify each instance endpoint separately, running one pganalyze collector Docker container for each instance:
```
msakrejda, line 60:
```suggestion
If you have multiple readers you want to monitor, you will need to specify each instance endpoint separately:
```
Though I guess we should update the example that follows to use the INI file syntax and have two separate sections. We can elide most of the config except the two `db_host` lines and the section names. We could use section names of `[primary]` and `[replica]` to make it clearer, too.

install/amazon_rds/04_configure_the_collector_ecs.mdx
```diff
@@ -99,21 +99,20 @@ Make sure to modify the values for `DB_HOST`, `DB_USERNAME`, `DB_NAME`, `AWS_REG
...
+Use the `cluster-ro` endpoint as the `DB_HOST` of the environment variables:
 
 ```json
 {"name": "DB_HOST", "value": "mydbcluster.cluster-ro-123456789012.us-east-1.rds.amazonaws.com"},
 ```
 
-For writer instances this is supported for any size of cluster, but for reader instances this is only supported in two-node clusters (i.e. single reader instance).
-
-If you have multiple readers you want to monitor you instead need to specify each instance endpoint separately, running one pganalyze collector Docker container for each instance:
+If you have multiple readers you want to monitor, you need to specify each instance endpoint separately, running one pganalyze collector Docker container for each instance:
```
msakrejda, line 115: As above, I think we should match changes in this doc with the other places that should mention `CONFIG_CONTENTS`. I guess we may want to centralize some of that information?

install/amazon_rds/04_configure_the_collector_docker.mdx
keiko713: Mentioning about `CONFIG_CONTENTS` here is a good idea, but I think that still can be a secondary option and we can suggest running two Docker containers as a primary option.
Or, do we actually want to push people to use `CONFIG_CONTENTS`?

msakrejda: Yeah, good point. I think running a single collector for multiple servers (especially when they're closely related like this) is more manageable, but the other way is a bit simpler. I'd lean to preferring `CONFIG_CONTENTS`, but no strong feelings.

install/amazon_rds/_04_setting_env_vars_multi.mdx
```diff
@@ -0,0 +1,40 @@
...
+  )
+}
+
+<CollectorDotEnv />
+
+This multiline variable definition is **only supported by Docker Compose and not by the Docker CLI**.
+If you're using the Docker CLI (e.g. `docker run --env-file pganalyze_collector.env`),
+please replace new lines with `\n`, like `CONFIG_CONTENTS="[pganalyze]\napi_key = ..."`.
+Alternatively, run one pganalyze collector Docker container for each server
+using the Single Server method.
```
keiko713, line 32: [note for reviewer] This is something I actually should be double checking by myself, but looking the following places, I think this is the current status.
https://github.com/containers/podman/issues/18724#issuecomment-1757162444
https://github.com/docker/cli/blob/master/opts/envfile.go

install/amazon_rds/_04_setting_env_vars_multi.mdx
```diff
@@ -0,0 +1,40 @@
...
+    <CodeBlock>
+      {`CONFIG_CONTENTS="[pganalyze]
+api_key = ${apiKey ?? 'your_pga_organization_api_key'}\n
+[server1]
+db_host = instance-id1.account-id.us-east-1.rds.amazonaws.com
+db_name = your_database_name
+db_username = your_monitoring_user
+db_password = your_monitoring_user_password
+db_sslrootcert = rds-ca-global
+db_sslmode = verify-full\n
```
msakrejda, line 14: Why do you have these escaped newlines only at the end of some lines?

msakrejda (approved):

install/amazon_rds/_04_setting_env_vars_multi.mdx
keiko713: Uhmm I tested this now but sadly this doesn't work. We're doing `echo "$CONFIG_CONTENTS"` but this doesn't handle new lines (`\n`)

https://github.com/pganalyze/collector/blob/32b62a13766663ed1167022cbfa41d2e1c89fca6/contrib/docker-entrypoint.sh#L6

```shell
$ CONFIG_CONTENTS="[pganalyze]\napi_key = test_key"
$ echo $CONFIG_CONTENTS 
[pganalyze]\napi_key = test_key
$ echo -e $CONFIG_CONTENTS 
[pganalyze]
api_key = test_key
```

There are two options:
1. add `-e` in the docker-entrypoint.sh and make a new release, then ship this doc
2. just say "If you're using the Docker CLI, it won't work, sorry"

... actually, this makes me wonder if Docker Compose way even works. With Docker Compose, we can define config var with new line without doing `\n` join, though at the end, what we do is `echo "$CONFIG_CONTENTS"` and if that part doesn't handle new line, I don't think it'll work 🤔

install/amazon_rds/_04_setting_env_vars_multi.mdx
keiko713: Good question, I kinda copy&pasted from the other place (EC2) but I was also wondering about this inconsistency too. Let me try to remove it and see if it still works.

keiko713: Ok I tested this, but looks like gatsby doesn't like an empty line (causes parse errors), so this workaround is needed.

PR #247, opened 2024-04-17
Consolidate per-database helper function instructions

msakrejda:
Currently, we prompt users to create helper functions to access column
stats and extended statistics separately, even though both of these
have to be created per-database.

Consolidate these, so users only have to iterate through their
databases once.

### DIFF
--- a/components/MonitoringUserSetupInstructions.tsx
+++ b/components/MonitoringUserSetupInstructions.tsx
@@ -69,6 +69,76 @@ $$ LANGUAGE sql VOLATILE SECURITY DEFINER;`}
   );
 };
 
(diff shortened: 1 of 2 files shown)


keiko713 (approved):
Looks good. Can you please update below as well to use a new `MonitoringUserPerDatabaseHelpers`?
https://github.com/pganalyze/pganalyze-docs/blob/deeda76648c284a46cf57612ddaee7139ced926c/install/crunchy_bridge/01_deploy_the_collector.mdx#L36-L38

msakrejda:
Ah, good catch. I looked for other uses of these with "Find all references" in VS Code, but did not find any--I'm not sure how I missed these.

PR #248, opened 2024-04-22
docs: set user id to 1000 on the ecs task

bmbferreira:
Minor improvement for the ecs documentation. If the user is not set to "1000", by default it will be the user with id 0 and this part of the collector entrypoint.sh will execute: https://github.com/pganalyze/collector/blob/main/contrib/docker-entrypoint.sh#L10-L12

This is a problem because it crashes on startup with the following error: 
```
/ $ setpriv --reuid=pganalyze --regid=pganalyze --inh-caps=-all --clear-groups
setpriv: unrecognized option: reuid=pganalyze
BusyBox v1.36.1 (2023-11-06 11:32:24 UTC) multi-call binary.

Usage: setpriv [OPTIONS] PROG ARGS

Run PROG with different privilege settings

-d,--dump               Show current capabilities
--nnp,--no-new-privs    Ignore setuid/setgid bits and file capabilities
--inh-caps CAP,CAP      Set inheritable capabilities
--ambient-caps CAP,CAP  Set ambient capabilities
```

### DIFF
--- a/install/amazon_rds/_04_setting_env_vars_multi_ecs.mdx
+++ b/install/amazon_rds/_04_setting_env_vars_multi_ecs.mdx
@@ -71,6 +71,7 @@ Now, save the following ECS task definition to a file, for example `pganalyze_ta
           "awslogs-stream-prefix": "pganalyze"
         }
       },
+      "user": 1000,
       "readonlyRootFilesystem": false,
       "mountPoints": []
     }
--- a/install/amazon_rds/_04_setting_env_vars_single_ecs.mdx
+++ b/install/amazon_rds/_04_setting_env_vars_single_ecs.mdx
@@ -66,6 +66,7 @@ Now, save the following ECS task definition to a file, for example `pganalyze_ta
           "awslogs-stream-prefix": "pganalyze"
         }
       },
+      "user": 1000,
       "readonlyRootFilesystem": false,
       "mountPoints": []
     }


lfittl:
> This is a problem because it crashes on startup with the following error:

Thanks for the report and the documentation patch!

I'm surprised this errors out regarding the `--reuid=pganalyze` flag, but I think there is no reason not to have ECS directly drop privileges, as that's what's also commonly done in Kubernetes.

@keiko713 Any conflicts this patch has with your current work around the setup guides? (I believe not, but wanted to double check before merging)

lfittl (approved):

keiko713:
Nice, thanks for the PR! It won't conflict with my current work, though I did create some duplicate docs around there in the previous change 🙈 (I should look into this more to avoid duplicates)
Could you please also update the following part too? (you modified `_04_setting_env_vars_multi_ecs.mdx` and below is for `_04_setting_env_vars_single_ecs.mdx`).

https://github.com/pganalyze/pganalyze-docs/blob/f163730f0e215fc417b8928f52200aeb9b7f40f5/install/amazon_rds/_04_setting_env_vars_single_ecs.mdx#L68-L70

bmbferreira:
@keiko713 done!

lfittl:
@bmbferreira Merged, thanks for the contribution!

I'll also kick off the internal process to get the website updated, that requires one more review and then it'll be live -- appreciate you bringing this up & providing a patch.

PR #252, opened 2024-04-25
Add AWS architecture diagram and note for cross account

keiko713:
Ok let's see if people like this. I like it overall, but one big downside is that now 00_overview page is almost too long.
I can also add "Collector can be installed in a different AWS account..." part in the diagram, like a footnote, which might work better (though maybe it's easier to review as a markdown :P).

image

### DIFF
--- a/install/amazon_rds/00_overview.mdx
+++ b/install/amazon_rds/00_overview.mdx
@@ -39,7 +39,7 @@ within your existing setup.
 
 <ImgCollectorAwsArchitecture />
 
-Collector can be installed in a different AWS account from the one hosting the
+The collector can be installed in a different AWS account from the one hosting the
 RDS/Aurora instance, provided that the collector can establish a connection to
 the RDS/Aurora instance (e.g., using VPC Peering).
 Additionally, the collector must be set up to assume an IAM role from the RDS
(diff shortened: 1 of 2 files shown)


install/amazon_rds/00_overview.mdx
```diff
@@ -31,6 +34,20 @@ ECS, or with Docker.
   * The collector instance needs to be able to connect to the server directly,
     bypass any connection poolers like PgBouncer
 
+Here is an overview diagram for how the pganalyze collector will be placed
+within your existing setup.
+
+<ImgCollectorAwsArchitecture />
+
+Collector can be installed in a different AWS account from the one hosting the
```
msakrejda, line 42: I guess this is somewhat subjective, but I think this would work better with the definite article:
```suggestion
The pganalyze collector can be installed in a different AWS account from the one hosting the
```
or just
```suggestion
The collector can be installed in a different AWS account from the one hosting the
```

msakrejda (approved):
Looks good. I wonder if it would work to move the "Proceed to Step 1" _above_ the diagram...

keiko713:
> I wonder if it would work to move the "Proceed to Step 1" above the diagram...

Yeah I thought about it but it looks slightly off to have a diagram after the button. I'll ship this version and happy to adjust if we feel that moving the button is better 👍

PR #255, opened 2024-05-02
Update stale links and point to auto explain

keiko713:
This has been bothering me quite a bit, let's update old links.

* Azure, Crunchy, Google => they have auto explain, so let's link to auto explain page
* Aiven => they actually don't support auto explain, so let's link to log based explain page

### DIFF
--- a/log-insights/setup/aiven/04_configure_service_integration.mdx
+++ b/log-insights/setup/aiven/04_configure_service_integration.mdx
@@ -72,5 +72,8 @@ show your integration as active:
   <p>
     We recommend setting up <a href="/docs/explain/setup/log_explain">Log-based EXPLAIN</a> as
     a follow-up step, to automatically EXPLAIN slow queries in Postgres.
+    As an alternative to Log-based EXPLAIN, you can integrate the auto_explain module with
+    <a href="https://github.com/aiven/aiven-extras?tab=readme-ov-file#examples">some changes</a> to
+    your application.
   </p>
 </PublicOnly>
--- a/log-insights/setup/aiven/04_configure_service_integration.mdx
+++ b/log-insights/setup/aiven/04_configure_service_integration.mdx
@@ -70,7 +70,7 @@ show your integration as active:
     <ImgLogInsightsScreenshot />
   </p>
   <p>
-    We recommend setting up <a href="/docs/explain/setup">Automated EXPLAIN</a> as
+    We recommend setting up <a href="/docs/explain/setup/log_explain">Log-based EXPLAIN</a> as
     a follow-up step, to automatically EXPLAIN slow queries in Postgres.
   </p>
 </PublicOnly>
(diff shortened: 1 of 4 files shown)


log-insights/setup/aiven/04_configure_service_integration.mdx
```diff
@@ -70,7 +70,7 @@ show your integration as active:
     <ImgLogInsightsScreenshot />
   </p>
   <p>
-    We recommend setting up <a href="/docs/explain/setup">Automated EXPLAIN</a> as
+    We recommend setting up <a href="/docs/explain/setup/log_explain">Log-based EXPLAIN</a> as
```
seanlinsley, line 73: Aiven actually does support auto_explain, but it's not well-documented and is unusual in that the `aiven_extras` extension must be created, and then auto_explain has to be enabled for the current database connection (it can't be enabled globally)

https://github.com/aiven/aiven-extras?tab=readme-ov-file#examples

```sql
-- Load the extension for the current session
SELECT * FROM aiven_extras.auto_explain_load();

-- Set minimum duration to 2000 ms
SELECT * FROM aiven_extras.set_auto_explain_log_min_duration('2000');

-- Enable ANALYZE
SELECT * FROM aiven_extras.set_auto_explain_log_analyze('on');

keiko713: According to https://github.com/aiven/aiven-extras/issues/16 , it can only be loaded for the active session. Uhmm I thought it's only enabled for the current psql session, but it's more like "session_preload_libraries" way? 🤔 If former, feels like it's not really usable.
Regardless, we don't currently have the Aiven documentation for how to enable it, so until we make that document, I think we can navigate people to Log-based EXPLAIN.

seanlinsley: It's definitely not ideal, but should be usable. It just means that an application has to run `auto_explain_load()` when opening a database connection.

msakrejda: Yeah, Lukas and I discussed this when we added Aiven and decided it's basically not worth the additional documentation complexity. No objection to revisiting that, but I think that's outside the scope of this PR.

msakrejda: That said, maybe it's worth saying something like

> We recommend and document our log-based EXPLAIN mechanism for Aiven, but it's possible to integrate the auto_explain module with [some changes to your application](https://github.com/aiven/aiven-extras?tab=readme-ov-file#examples).

?

(Linking to the example Sean shared.)

keiko713: I added something similar to "it's possible to integrate the auto_explain module with some changes to your application", though I changed a little bit as Maciek's suggestion is a bit lacking "do this as a follow-up step" aspect.

lfittl (approved):

PR #256, opened 2024-05-07
Enterprise Server release 2024.04.2

lfittl:


keiko713 (approved):

seanlinsley (approved):
This patch release only includes the cherry-picked commits for the problems covered in the bugfixes section, so there aren't any security updates like mentioned in the changelog, but otherwise this looks good.

lfittl:
> This patch release only includes the cherry-picked commits for the problems covered in the bugfixes section, so there aren't any security updates like mentioned in the changelog, but otherwise this looks good.

The security updates on the base image would still apply (I omitted the libraries part of that text) since that's essentially just what an `apt-get upgrade` pulls in on the image rebuild.

PR #260, opened 2024-05-17
Collector settings: Clarify that DB_ALL_NAMES uses 0/1

lfittl:
Whilst we also do support "false" as is indicated in the documentation, we generally prefer the short-hand "0" or "1" for boolean flags.

### DIFF
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -178,9 +178,9 @@ helper functions are expected to be defined, and the one we connect to for serve
     </tr>
     <tr>
       <td><code>DB_ALL_NAMES</code></td>
-      <td>false</td>
+      <td>0</td>
       <td>Alternative to setting <code>*</code> as the last entry of db_name (<code>DB_NAME</code>) to
-      monitor all databases on the server. Can only be set using an environment variable.<br />
+      monitor all databases on the server. Can only be set using an environment variable (<code>DB_ALL_NAMES = 1</code>).<br />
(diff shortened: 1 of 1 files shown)


keiko713 (approved):

msakrejda (commented):
Why prefer 0/1? I always found 0/1 for boolean options to be forcing the user to speak the computer's language.

msakrejda:
In addition, other boolean settings in this file use `false` as the default.

lfittl:
> In addition, other boolean settings in this file use `false` as the default.

That's a good point - let me rethink this. Maybe we should consistently use true/false instead (I think some of the provider-specific guides use 1/0 currently).

seanlinsley:
Maybe we could go ahead and merge this PR to make the docs more accurate, then make it a future work item to standardize the boolean envvars? Sidenote: the best way to avoid backwards compatibility issues might be to make them all accept both 0/1 and true/false, using a helper method to parse the value.

lfittl:
> Sidenote: the best way to avoid backwards compatibility issues might be to make them all accept both 0/1 and true/false, using a helper method to parse the value.

I think we already do this, no?

I was thinking of modifying things to consistently be true/false, per Maciek's note - I don't think there is a good reason to require 0/1 really.

lfittl:
As discussed we don't need to modify any existing examples to be consistent in using `true`/`false`, so I'll close this.

PR #263, opened 2024-06-14
Add an example of using k8s secret with EKS setup

keiko713:
Currently, this part is pretty hand wavy as it just says "follow your organization's security best practices and use Kubernetes secrets". With this, I mention how it actually can be used.
For the k8s secret, the key must consist of alphanumeric characters, `-`, `_` or `.`, so we should be able to use `pganalyze-collector.conf` as the key.

Reference:
https://kubernetes.io/docs/concepts/configuration/secret/#use-case-dotfiles-in-a-secret-volume
https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data

### DIFF
--- a/install/amazon_rds/04_configure_the_collector_eks.mdx
+++ b/install/amazon_rds/04_configure_the_collector_eks.mdx
@@ -50,6 +50,39 @@ best practices and use Kubernetes secrets.
 
 <SettingEnvVarsEKS apiKey={props.apiKey} />
 
+When using Kubernetes secrets, it's easier to use the `INI` file style for the
+config. You can find the example of this in the Multiple DB Instances example
+above, where it is used as the value of the `CONFIG_CONTENTS` environment variable.
+
+For instance, you can create a secret called `pganalyze-secrets` and add a key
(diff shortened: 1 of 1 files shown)


msakrejda (approved):

install/amazon_rds/04_configure_the_collector_eks.mdx
```diff
@@ -50,6 +50,39 @@ best practices and use Kubernetes secrets.
 
 <SettingEnvVarsEKS apiKey={props.apiKey} />
 
+When using Kubernetes secrets, it's easier to use the `INI` file style for the
+config. You can find the example of this in the Multiple DB Instances example
+above, where it is used as the value of the `CONFIG_CONTENTS` environment variable.
+
+For instance, you can create a secret called `pganalyze-secrets` and add a key
+`pganalyze-collector.conf`, setting the `INI` collector config as its value.
```
keiko713, line 58: Re-reading this part, I think it's the best to have an example yaml file so I'm going to actually add it.

PR #265, opened 2024-06-25
Add a simple collector uninstalling step docs

keiko713:
It's super simple one, but good enough as a starter

### DIFF
--- a/collector/uninstalling.mdx
+++ b/collector/uninstalling.mdx
@@ -1,0 +1,84 @@
+---
+title: 'Uninstalling the pganalyze Collector'
+backlink_href: /docs/collector
+backlink_title: 'pganalyze Collector'
+---
+
+To uninstall pganalyze from your system, follow these two steps:
+
+1. [Cleaning up a monitoring user and helper functions](#cleaning-up-a-monitoring-user-and-helper-functions)
+2. [Uninstalling the collector](#uninstalling-the-collector)
+
+## Cleaning up a monitoring user and helper functions
+
+To uninstall the pganalyze collector, first you'll need to clean up a pganalyze
+monitoring user and helper functions that you created during the setup.
+
+* Run `DROP SCHEMA pganalyze CASCADE` in each database that you monitor to
+  remove helper functions
+* Run `DROP USER pganalyze` in any database to remove a monitoring user
+
+
+## Uninstalling the collector
+
+Since there are many different ways to run the collector, the uninstall process
+will depend on how you originally set up the collector:
+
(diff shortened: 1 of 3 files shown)


collector/uninstalling.mdx
```diff
@@ -0,0 +1,84 @@
...
+```
+sudo yum remove pganalyze-collector
+```
+
+Then, clean up a state file and its directory:
+```
+sudo rm -r /var/lib/pganalyze-collector
+```
+
+Finally, clean up the user and group:
```
msakrejda, line 48:
```suggestion
Finally, clean up the collector system user and group:
```
I think clarifying "system" user here makes it clearer what you're doing. It should be pretty obvious from the `DROP USER` above, but maybe the reader skipped past that section and jumped right into this.

collector/uninstalling.mdx
```diff
@@ -0,0 +1,84 @@
...
+On systems using Debian-based packages, you can uninstall the collector through
+the standard uninstall mechanism.
+
+First, remove the collector package from your system (this includes disabling
+and stopping the service):
+```
+sudo apt remove pganalyze-collector
+```
+
+Then, clean up a state file and its directory:
```
msakrejda, line 64: Same as above.

collector/uninstalling.mdx
```diff
@@ -0,0 +1,84 @@
...
+```
+sudo rm -r /var/lib/pganalyze-collector
+```
+
+Finally, clean up the user and group:
```
msakrejda, line 69: Same as above.

msakrejda (approved):
Nice, this is great. I suggested "remove" or "drop" in some places, since "clean up" is a little vague, and when you're doing a specific action, we can talk about it more concretely.

collector/uninstalling.mdx
```diff
@@ -0,0 +1,84 @@
...
+2. [Uninstalling the collector](#uninstalling-the-collector)
+
+## Cleaning up a monitoring user and helper functions
+
+To uninstall the pganalyze collector, first you'll need to clean up a pganalyze
+monitoring user and helper functions that you created during the setup.
+
+* Run `DROP SCHEMA pganalyze CASCADE` in each database that you monitor to
+  remove helper functions
+* Run `DROP USER pganalyze` in any database to remove a monitoring user
```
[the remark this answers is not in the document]
keiko713: Yeah exactly. I think there are other things that are customizable (e.g. the location of the config/state file, etc.) but I didn't call them out here as I wanted to keep this doc simple.

PR #270, opened 2024-07-24
Collector settings: Clarify the allowed values for boolean settings

lfittl:


keiko713 (approved):
Can you update here too?
https://github.com/pganalyze/pganalyze-docs/blob/e97dc5c34825e01a45026500f153ac3c87a94684/install/crunchy_bridge/01_deploy_the_collector.mdx#L14

PR #271, opened 2024-07-25
Add Index Consolidation docs

msakrejda:


index-advisor/consolidation.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+## Limitations
+
+ - Dropping the wrong index can have a catastrophic impact on database
+   performance. We recommend [testing](/docs/index-advisor/test-insights) all
+   Index Advisor insights before implementing them, but this is especially true
+   of consolidation insights.
+ - Index Advisor currently has a limited model of partial indexes (indexes with
+   a `WHERE` clause), and will likely recommend dropping them: these insights
+   should be ignored, or at least reviewed very critically.
```
seanlinsley, line 32: I feel like this is under-selling the impact by being number 2 in a list, and the use of vague terminology ("will likely" instead of "will"). How about:
> Index Advisor doesn't generate index recommendations using `WHERE` clauses because of technical limitations, so when index consolidation is turned on it will recommend dropping all partial indexes. This guidance should be disregarded.

Another option might be to automatically hide removal recommendations of partial indexes in the app, to avoid the potential footgun from team members who haven't read this document or have forgotten the details.

msakrejda: Yeah, I think your stronger wording makes sense. I'll update.

We discussed hiding these in-app and I think that makes sense.

seanlinsley (approved):
This seems reasonable, though others might want to weigh in.

jensy (approved):
I think this is good! Thanks for writing this up.

msakrejda:
Given the two approvals, I'll merge now so we have some docs to point customers to, but I'm happy to update this further if anyone else has thoughts on this.

PR #274, opened 2024-08-26
Include expression indexes in Index Unused check

seanlinsley:
Screenshot 2024-08-26 at 2 58 09 PM

### DIFF
--- a/components/CheckDocumentation/schema/IndexUnused.tsx
+++ b/components/CheckDocumentation/schema/IndexUnused.tsx
@@ -27,9 +27,8 @@ const IndexUnusedTrigger: React.FunctionComponent<CheckTriggerProps> = ({
         periods.
       </p>
       <p>
-        Ignores small indexes (less than 32kB), as well as primary keys, unique
-        indexes, and expression indexes since they are necessary to enforce
-        constraints or useful as query planning hints.
+        Ignores small indexes (less than 32kB), primary keys, and unique indexes
+        since they are necessary to enforce constraints.
       </p>
     </>
   );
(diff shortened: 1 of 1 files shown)


components/CheckDocumentation/schema/IndexUnused.tsx
```diff
@@ -69,6 +68,11 @@ const IndexUnusedGuidance: React.FunctionComponent<CheckGuidanceProps> = ({
           </>
         )}
       </p>
+      <p>
+        Note: when deleting expresssion indexes, you may want to run
+        {" "}<code>CREATE STATISTICS</code> to have better query plans that are
+        aware of the expression results.
+      </p>
```
msakrejda, line 75: I think it would be nice to give more concrete advice here. "You may want to" is technically correct, but leaves the user wondering whether they do or do not want to. Given this is a weird edge case and not commonly understood, I don't think we need to say anything super-specific, but we could share what we know.

So something like,

```tsx
<p>
  Note: Unused expression indexes can still influence query planner statistics and
  affect query planning. When dropping the index, monitor performance and use
  <PGDocsLink path="sql-createstatistics.html">CREATE STATISTICS</PGDocsLink>
  with the expression from this index if necessary.
</p>
```

I guess we could even include `CREATE STATISTICS` DDL (at least in-app) if we really wanted to, but that seems like overkill.

Thoughts?

seanlinsley: I've updated it to this:
>Note: expression indexes can influence the query planner even if the index is unused. When dropping an expression index, if you notice some queries are slower you may want to [CREATE STATISTICS](https://www.postgresql.org/docs/current/sql-createstatistics.html) for the expression the index previously covered.

msakrejda (approved):

PR #276, opened 2024-08-27
Add paragraph for unsupported providers

jensy:


index.mdx
```diff
@@ -16,6 +16,8 @@ To continue, select in which environment you are running your Postgres database:
 
 <OverviewInstallChoice />
 
+If you're interested in integrating pganalyze with your database hosting provider but don't see it listed as officially supported, you can follow the [installation guide for self-managed databases](/docs/install/self_managed/00_choose_setup_method). You will need to run the pganalyze collector on a server that can reach your database in order to connect it to your provider. Note that [Log Insights](/docs/log-insights) and other features that depend on it will have some limitations.
```
msakrejda, line 19: A couple of notes:
 - changed link to manual self-managed instructions: the guided setup instructions won't work in this case, so we might as well save users a step
 - reworded connectivity requirement a bit
 - noted that system stats are not available
 - noted that Log Insights _is_ available if the provider offers Postgres logs via syslog

```suggestion
If you're interested in integrating pganalyze with your database hosting provider but don't see it listed as officially supported, you can follow the [manual installation guide for self-managed databases](docs/install/self_managed/01_create_monitoring_user). You will need to run the pganalyze collector on a server with network connectivity to your database in order to monitor your database. Note that system-level statistics (memory usage, I/O, etc.) cannot be collected this way and will not be available in pganalyze. [Log Insights](/docs/log-insights) (and the features that depend on it) is only available if your provider exports Postgres logs for your database server [via syslog](https://pganalyze.com/docs/log-insights/setup/syslog-server).
```

Does that make sense?

jensy: Totally!

msakrejda: Great, but the other text should be updated to match: I only created a suggestion for this one.

PR #278, opened 2024-08-29
Fix Log Insights limits settings docs

msakrejda:
This doc updated was missed in https://github.com/pganalyze/collector/pull/301


keiko713 (approved):
Nice 👍 I wonder if we could link this from the app potentially, as a hint to users who may not see logs due to this.

msakrejda:
Yeah, that might be nice.

PR #279, opened 2024-09-04
Update EKS setup docs to include Pod Identity usage

keiko713:
With a new collector, now people can use Pod Identity with EKS. I double checked and confirmed that this works with a new collector.
Revive the doc that I created previously to mention the Pod Identity way.

image

### DIFF
--- a/install/amazon_rds/04_configure_the_collector_eks.mdx
+++ b/install/amazon_rds/04_configure_the_collector_eks.mdx
@@ -29,7 +29,8 @@ helm show values pganalyze/pganalyze-collector > values.yaml
 
 Update the serviceAccount section in the values.yaml file to use the role just
 created in the previous step. For the name, use "pganalyze-service-account" as
-specified in the assume role policy:
+specified in the Pod Identity association (for Pod Identity) or the assume role
+policy (for IRSA):
 
 ```yaml
 serviceAccount:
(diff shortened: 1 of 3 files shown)


install/amazon_rds/04_configure_the_collector_eks.mdx
```diff
@@ -29,14 +29,16 @@ helm show values pganalyze/pganalyze-collector > values.yaml
...
+policy (for IRSA):
 
 ```yaml
 serviceAccount:
   # -- Specifies whether a service account should be created
   create: true
   # -- Annotations to add to the service account
   annotations:
+    # Add this annotation only if you're using IRSA (not required for Pod Identity)
     eks.amazonaws.com/role-arn: "arn:aws:iam::<aws-account-id>:role/pganalyzeServiceAccountRole"
```
msakrejda, line 42: If Pod Identity is our preferred mechanism, should we have this commented-out by default? No strong feelings, just thought I'd bring it up.

msakrejda (approved):

install/amazon_rds/04_configure_the_collector_eks.mdx
keiko713: Great point, actually. I think I'll change this to below 👍

```
  # -- Annotations to add to the service account
  annotations: {}
  # Update above annotation if you're using IRSA
  # annotations:
  #   eks.amazonaws.com/role-arn: "arn:aws:iam::<aws-account-id>:role/pganalyzeServiceAccountRole"
```

PR #281, opened 2024-09-11
Enterprise Server release 2024.04.4

lfittl:


enterprise/releases/2024-04-4.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+## Notes
+
+- This is a security patch release on top of **[2024.04.0](2024-04-0)**, **[2024.04.1](2024-04-1)**, **[2024.04.2](2024-04-2)** and **[2024.04.3](2024-04-3)**
+  - This release does not have any new database migrations, and is easy to apply (or rollback)
+  - If you're upgrading from a release before [2024.04.0](2024-04-0), be sure to read those release notes,
+    and follow the [recommended upgrade steps](/docs/enterprise/upgrade)
+- We recommend upgrading as soon as possible if you are utilizing SAML authentication in pganalyze
+  - **Security impact:** The resolved [CVE-2024-45409](https://www.cve.org/CVERecord?id=CVE-2024-45409) in ruby-saml would have allowed
+    impersonating any member within your pganalyze organization by an internal threat actor with
```
keiko713, line 17: I think it's fine to leave it as is, but just as a note, the attacker doesn't need to be an internal threat actor, as long as they managed to get a SAML response somehow (e.g. through social engineering, etc.). However, an internal threat actor is much easier to do this (also very feasible), also if we talk about "you can get this important info with social engineering", we can talk about "you can leak your password to the attacker!" level, so I think it's okay to only mention this as an internal threat actor.

keiko713 (approved):

enterprise/releases/2024-04-4.mdx
lfittl: Yeah, I think you would have to leak something that could be considered a credential (a signed SAML assertion for an application you have access to) in order for an external party to exploit this, so I think this wording is fine to keep as is.

PR #284, opened 2024-09-26
Clarify how to monitor Aurora cluster with multiple readers

keiko713:
Maybe it's only me with my weird reading, but when you say "If you have multiple readers you want to monitor, do X", I feel like "well I don't need to monitor them all, let me skip reading/doing X" even though I have multiple readers with a cluster.
Actually, I was always wondering what these sentences actually means, and I didn't fully understand this until today. So I think it's the best to remove any ambiguity and clarify that monitoring them individually is necessary (IOW, using  for Aurora cluster with multiple readers.

Actually, just noticed that EC2 doc does say "this is only supported in two-node clusters (i.e. single reader instance)" for `cluster-ro` way, which is really clear. We could adopt this to other docs too 🤔

### DIFF
--- a/install/amazon_rds/04_configure_the_collector_docker.mdx
+++ b/install/amazon_rds/04_configure_the_collector_docker.mdx
@@ -47,9 +47,9 @@ Create a new configuration file (e.g. named `pganalyze_collector.replica.env`) a
 DB_HOST=mydbcluster.cluster-ro-123456789012.us-east-1.rds.amazonaws.com
 ```
 
-If you have multiple readers you want to monitor, you either need to add it to
-the content of `CONFIG_CONTENTS`, or run one pganalyze collector Docker
-container for each instance.
+If you have multiple readers with a cluster, you either need to add each instance
+endpoint to the content of `CONFIG_CONTENTS` (e.g. add each sections like `[reader_instance1]`, `[reader_instance2]`),
+or run one pganalyze collector Docker container for each instance.
 
 ## Test snapshot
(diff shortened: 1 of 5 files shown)


seanlinsley (approved):

keiko713:
Completely forgot to merge this.
Re-reading, I'm not sure if Keiko back in September is having a point or not (the pre-fix version reads actually okay), but let's apply this change and see how it goes, we can always revisit if this version is not so good.

install/amazon_rds/04_configure_the_collector_docker.mdx
```diff
@@ -47,9 +47,9 @@ Create a new configuration file (e.g. named `pganalyze_collector.replica.env`) a
 DB_HOST=mydbcluster.cluster-ro-123456789012.us-east-1.rds.amazonaws.com
 ```
 
-If you have multiple readers you want to monitor, you either need to add it to
-the content of `CONFIG_CONTENTS`, or run one pganalyze collector Docker
-container for each instance.
+If you have multiple readers with a cluster, you either need to add each instance
```
lfittl, line 50: @keiko713 The first part of that sentence reads a bit odd to me - maybe the following would be better?

"If you have a cluster with multiple readers, you either ..."

(same revision in the other files changed)

keiko713: nice, yeah "a cluster with multiple readers" reads better, will update 👍

PR #285, opened 2024-10-09
Update get_column_stats for Postgres 17

seanlinsley:
On Postgres 17, the function breaks:

```sql
select * from pganalyze.get_column_stats() limit 1;
ERROR:  return type mismatch in function declared to return pg_stats
DETAIL:  Final statement returns too few columns.
CONTEXT:  SQL function "get_column_stats" during startup
```

Note that this documentation update makes it so this function definition no longer works for previous versions of Postgres. Do we have a UI pattern to handle this, like a tabbed component that lets you switch between PG 16 and 17 versions of the function?

This error wouldn't have occurred if the function didn't `RETURNS SETOF pg_stats`, instead choosing to return its own row type. If we don't plan to use the new fields, maybe we should change the return type instead of adding the new fields?

Here's the documentation for the new fields. IIRC this function is intended to return all data points that are not considered sensitive (don't include user data), so I've chosen to return null for `range_bounds_histogram`

Screenshot 2024-10-09 at 11 05 37 AM

### DIFF
--- a/components/MonitoringUserSetupInstructions.tsx
+++ b/components/MonitoringUserSetupInstructions.tsx
@@ -41,7 +41,7 @@ CREATE OR REPLACE FUNCTION pganalyze.get_column_stats() RETURNS SETOF pg_stats A
 $$
   /* pganalyze-collector */ SELECT schemaname, tablename, attname, inherited, null_frac, avg_width,
     n_distinct, NULL::anyarray, most_common_freqs, NULL::anyarray, correlation, NULL::anyarray,
-    most_common_elem_freqs, elem_count_histogram
+    most_common_elem_freqs, elem_count_histogram, range_length_histogram, range_empty_frac, NULL::anyarray
   FROM pg_catalog.pg_stats
   WHERE schemaname NOT IN ('pg_catalog', 'information_schema') AND tablename <> 'pg_subscription';
 $$ LANGUAGE sql VOLATILE SECURITY DEFINER;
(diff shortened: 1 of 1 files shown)


lfittl:
> This error wouldn't have occurred if the function didn't `RETURNS SETOF pg_stats`, instead choosing to return its own row type. If we don't plan to use the new fields, maybe we should change the return type instead of adding the new fields?

Yeah, that seems reasonable, and is how we solved this for `pg_stats_ext`.

@msakrejda Any concerns with changing this? (AFAIK you were involved in the original discussion on this for extended stats with Keiko and myself, though maybe I'm misremembering)

msakrejda:
Looks like that was in #217--that doesn't seem to have any discussion on using the raw type versus TABLE, but I think TABLE makes more sense. +1 on changing pg_stats to do that until we need those fields.

And for what it's worth, we do have a way of doing version-specific instructions. #135 dropped that since that dropped 9.6 support and all supported versions could use identical instructions, but we can bring it back if necessary.

seanlinsley:
How do we communicate this to everyone without it turning into a support ticket when they upgrade to Postgres 17? It's not even as simple as pointing them to the documentation since the function signature changed, so `CREATE OR REPLACE FUNCTION` doesn't work; they'll need to drop the existing function.

msakrejda:
Oh, also, these are duplicated [here](https://github.com/pganalyze/collector/blob/main/runner/generate_helper_sql.go#L16-L24) :frowning:.

msakrejda:
Re: the upgrade problem—I suppose we could have the collector recognize the old function definition at startup and print instructions?

msakrejda (approved):

LukasJK:
I think the troubleshooting pages still need updating?

PR #286, opened 2024-10-10
Remove object storage from enterprise documentation

seanlinsley:
The upcoming enterprise release won't require object storage, so this PR removes that documentation to avoid confusion.


enterprise/setup/index.mdx
```diff
@@ -26,7 +26,7 @@ The pganalyze container image is mostly self-contained, and runs various service
   * 50 GB of dedicated disk space or more
 * Optional: If you use the RDS integration, set up an appropriate IAM role that can be used to download logs and system metrics
   * You can also assign the required policy to an existing EC2 instance and its instance role
-  * You will need both the [collector IAM policy](/docs/install/amazon_rds/iam_policy) as well as the [object storage IAM policy](/docs/enterprise/setup/object-storage-s3)
+  * You will nned the [collector IAM policy](/docs/install/amazon_rds/iam_policy)
```
keiko713, line 29:
```suggestion
  * You will need the [collector IAM policy](/docs/install/amazon_rds/iam_policy)
```

keiko713:
Overall looking good!
I wonder if we should add some note for the existing customers:
* the breaking change of the collector (can't use with 0.60.0 with old Enterprise release)
* people can keep having object storage (but won't be used once all collectors are 0.60.0+ && after 7 days of upgrade)
* some other things that might be worth noting (e.g. the context, the old Enterprise had the thing called object storage and had to setup but no longer needed)

It doesn't need to have a dedicated page, uhmm, maybe in troubleshooting page or something.

seanlinsley:
I was expecting that to be explained in the changelog. I'm not sure we need anything more than that.

keiko713 (approved):
Ah good point, if that is going to be explained in the changelog, that's indeed enough.

Are we going to merge this before writing changelog (aka release enterprise) though? I guess it's unavoidable given that you want to include this in the app PR
We just need to be careful that we won't release this to the web before the enterprise release

PR #293, opened 2024-12-03
Add Tune Queries as an option of DEFAULT_ORG_ROLE

keiko713:
For a new permission - don't need to merge yet but creating a PR before forget.


msakrejda (approved):

keiko713:
I included the Tune Queries section to the permission page too in this PR. I paired with @lfittl  about this change, so I'll go ahead and merge this.

PR #294, opened 2024-12-03
Add Buffer Cache documentation

seanlinsley:


schema-statistics/buffer-cache.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+If the collector runs into permissions issues when querying pg_buffercache, run:
+
+```sql
+GRANT USAGE ON SCHEMA public TO pganalyze;
+GRANT SELECT ON pg_buffercache TO pganalyze;
+GRANT EXECUTE ON FUNCTION pg_buffercache_pages TO pganalyze;
+```
+
+Note that by default this is not monitored for servers with a buffer cache larger than 200 GB. To change this, set `max_buffer_cache_monitoring_gb = 300` in the collector config.
```
lfittl, line 27:
```suggestion
## Performance implications of collecting shared buffer statistics

Since Postgres 10, the pg_buffercache view uses reduced locking and as such is generally safe to query on production systems without impacting the regular workload. pganalyze queries the buffer cache data every 10 minutes.

By default this feature is only turned on for servers with `shared_buffers` configured for 200 GB of RAM or less. You can change the threshold by setting the `max_buffer_cache_monitoring_gb` setting, e.g. to `300` in the collector config.

This threshold exists because `pg_buffercache` incurs a fixed CPU cost for each buffer page (more buffer pages = more CPU used by querying it), as well as the fact that it temporarily writes its data to a temporary file, which has a size of roughly 0.5% of the `shared_buffers` value, or about 1GB at 200 GB `shared_buffers`.
```

schema-statistics/buffer-cache.mdx
```diff
@@ -0,0 +1,46 @@
...
+title: 'Tracking Postgres buffer cache usage in pganalyze'
+backlink_href: /docs/schema-statistics
+backlink_title: 'Tracking Postgres schema statistics in pganalyze'
+---
+
+Postgres caches recently-used table and index data in its buffer cache, whose size is configured by the `shared_buffers` setting. pganalyze can track buffer cache usage with the [pg_buffercache](https://www.postgresql.org/docs/current/pgbuffercache.html) extension, helping you to understand database access patterns over time and identify tables that may need to be optimized.
+
+## Setup
+
+Using collector version 0.63.0 or later, enable the extension:
```
lfittl, line 11:
```suggestion
Using collector version [0.63.0 or later](/docs/collector/upgrading), enable the extension:
```

lfittl (approved):
Feel free to merge once comments addressed and I can update the screenshots later to take into account the font size change on the headings.

PR #297, opened 2024-12-26
Update IAM policy link to access from the app

keiko713:
These two docs can be rendered within the app, and without the absolute path, the link will be broken (or rather, page won't be loaded).
Update the link to the full path so that it can be opened as a docs page in the separate window when accessed from the app.

### DIFF
--- a/api/index.mdx
+++ b/api/index.mdx
@@ -26,6 +26,7 @@ change pganalyze settings.
 
 - [Creating an API key](/docs/api/create-api-key)
 - [Queries](/docs/api/queries)
+    * [getServers - Get a list of servers for an organization](/docs/api/queries/getServers)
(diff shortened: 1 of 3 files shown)


api/index.mdx
```diff
@@ -26,6 +26,7 @@ change pganalyze settings.
 
 - [Creating an API key](/docs/api/create-api-key)
 - [Queries](/docs/api/queries)
+    * [getServers - Get a list of servers for an organization](/docs/api/queries/getServers)
```
keiko713, line 29: sorry this is unrelated but just noticed that it's not listed in https://pganalyze.com/docs/api so I wanted to add it

lfittl (approved):

PR #299, opened 2025-01-04
Add description of how to run the test with EKS and ECS

keiko713:
We don't show how to run the test command with EKS setup well, so let's add it.
I've added ECS too. Haven't tested with ECS but it should... work?


lfittl (approved):
:+1:

PR #305, opened 2025-02-06
Update Google Cloud SQL doc to support IAM Authentication

keiko713:


install/google_cloud_sql/_01_create_monitoring_user_iam.mdx
```diff
@@ -0,0 +1,48 @@
...
+### Configure new and existing instances for IAM database authentication
+
+Once the service account for a monitoring user is created, configure the database to monitor for IAM database authentication.
+Follow the [Cloud SQL documentation](https://cloud.google.com/sql/docs/postgres/create-edit-iam-instances) to configure.
+
+You can skip this part if your database is already configured for IAM database authentication.
+
+### Add a service account to a Cloud SQL instance
+
+Next, add a service account to a Cloud SQL instance to monitor. This is a step to create an actual Postgres user using this service account.
```
msakrejda, line 29:
```suggestion
Next, add a service account to a Cloud SQL instance to monitor. This creates an actual Postgres user for this service account.
```

install/google_cloud_sql/_01_create_monitoring_user_iam.mdx
```diff
@@ -0,0 +1,48 @@
...
+Following is the steps to add using the Google Cloud console. Please refer the
```
msakrejda, line 30:
```suggestion
The following steps add the service account using the Google Cloud console. Please refer to the
```

install/google_cloud_sql/_01_create_monitoring_user_iam.mdx
```diff
@@ -0,0 +1,48 @@
...
+[Cloud SQL documentation](https://cloud.google.com/sql/docs/postgres/add-manage-iam-users#creating-a-database-user)
+for the other ways to add it.
```
msakrejda, line 32:
```suggestion
for other ways to add it.
```

install/google_cloud_sql/_01_create_monitoring_user_psql.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+Note that it is important you run these as a user with **cloudsqlsuperuser / alloydbsuperuser** privileges in order to pass down the full access to statistics tables.
+
+## Verify connection
+
+Let's make sure we can connect to the database with the new user:
+
+<VerifyConnection host="1.2.3.4" />
+
+Write down the username and password of the monitoring user, we will use it in the last step of this tutorial.
```
msakrejda, line 18:
```suggestion
Write down the username and password of the monitoring user, we will use it in the last step of this guide.
```

msakrejda (approved):
I left some comments regarding wording, but in general this looks good. I have not walked through it.

install/google_cloud_sql/01_create_monitoring_user.mdx
```diff
@@ -5,8 +5,31 @@ backlink_href: /docs/install
...
+  ];
+  return (
+    <TabPanel items={tabs}>
+      {(idx) => {
+        const tab = tabs[idx][0];
+        switch (tab) {
+          case 'psql':
+            return <TabTextContent><CreateMonitoringUserPsql /></TabTextContent>;
+          case 'iam':
+            return <TabTextContent><CreateMonitoringUserIam /></TabTextContent>;
```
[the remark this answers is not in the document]
keiko713: I'm actually not sure, @lfittl do we expect people to use this mainly moving forward?

install/google_cloud_sql/_01_create_monitoring_user_iam.mdx
```diff
@@ -0,0 +1,48 @@
...
+This means you can use a service account as a monitoring user.
+
+First, create a new service account with the ID "pganalyze", and set the following two roles:
+* **Cloud SQL Client** (`roles/cloudsql.client`)
+  * Provides connectivity access to Cloud SQL instances
+* **Cloud SQL Instance User** (`roles/cloudsql.instanceUser`)
+  * Role allowing access to a Cloud SQL instance
+
+The email address `pganalyze@your-project-id.iam.gserviceaccount.com` will be given to this service account.
+For the database username, Cloud SQL for PostgreSQL truncates the `.gserviceaccount.com` suffix in the email,
```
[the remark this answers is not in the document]
keiko713: I agree with you, so this is simply FYI (I'll update this with your suggestion) but I copy&pasted things from Google's doc and they used "truncates" for this one:
https://cloud.google.com/sql/docs/postgres/add-manage-iam-users#creating-a-database-user

collector/settings.mdx
```diff
@@ -335,7 +335,7 @@ Note that the `aws_endpoint_*` settings are only relevant if you are using custo
     </tr>
     <tr>
       <td>db_use_iam_auth (<code>DB_USE_IAM_AUTH</code>)</td>
-      <td>[none]</td>
+      <td>false</td>
```
keiko713, line 338: Side note (maybe for the future), but we should make up our mind about how to say "there is no default value".

We use `[none]` or `n/a` here and there, and for the case like this, while "[none]" is correct (in the way that the golang treats that as boolean "false") but it's not so clear from here.

install/google_cloud_sql/_01_create_monitoring_user_psql.mdx
keiko713: I was wondering about this word "tutorial" as I don't really remember that I was writing down this.
Turned out that we actually use this sentence in the other places, and I copy & pasted from there, that's why it's there.
I'll go ahead and update places that say "tutorial" too.

install/google_cloud_sql/01_create_monitoring_user.mdx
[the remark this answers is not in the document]
keiko713: I'm going to merge as is for now, but happy to swap position as a follow-up PR

collector/settings.mdx
msakrejda: Ah, my bad for not making this clearer. I originally started using `[none]` as "this has no default but is ignored if you don't set it" and `n/a` as "this has no default, it must be configured (for a certain hosting provider or hosting configuration)".

This wasn't a clear delineation to begin with, because arguably `db_sslrootcert_contents` in the first commit should have been `[none]` rather than `n/a`, and I'm sure over time we've added more settings that don't follow this pattern (since I didn't make it clear), but that was the idea. I think your suggested change here makes sense, though.

install/google_cloud_sql/01_create_monitoring_user.mdx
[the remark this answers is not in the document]
lfittl: I think leaving it second is better - IAM authentication is a good security best practice, but I would still only expect 10-20% of people to use it.

(and if its anything like on AWS, the extra steps in enabling it for the database server mean people would usually not do this just for setting up pganalyze, if they don't already use it elsewhere)

collector/settings.mdx
keiko713: I see, thanks for explaining. Yeah, I think it'd be good to note it (can even live in this docs page too), and enforce the usage.
Let me see if I can make a PR for this 👍

install/google_cloud_sql/01_create_monitoring_user.mdx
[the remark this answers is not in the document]
keiko713: Yeah that's my feeling too, will leave this as is 👍

PR #307, opened 2025-02-15
Flesh out PII filtering docs

msakrejda:
Maybe this should be broken out to a separate section? It would
probably also be useful to note what parts of the application are
impacted, but I left that as a follow-up step (possibly still in this
patch).

### DIFF
--- a/collector/pii-filtering.mdx
+++ b/collector/pii-filtering.mdx
@@ -1,7 +1,7 @@
 ---
 title: 'PII Filtering'
-backlink_href: /docs
-backlink_title: 'Documentation'
+backlink_href: /docs/collector
+backlink_title: 'pganalyze Collector'
 ---
 
 [Personally Identifiable
--- a/directory.json
+++ b/directory.json
@@ -88,6 +88,10 @@
       "title": "pganalyze Collector packages",
       "path": "/docs/collector/packages"
     },
+    "collector/pii-filtering": {
+      "title": "PII Filtering",
+      "path": "/docs/collector/pii-filtering"
+    },
     "collector/settings": {
       "title": "pganalyze Collector settings",
       "path": "/docs/collector/settings"
@@ -999,10 +1003,6 @@
     "opentelemetry/traceparent": {
       "title": "Set up traceparent query tag within your application",
       "path": "/docs/opentelemetry/traceparent"
-    },
-    "pii-filtering": {
-      "title": "PII Filtering",
-      "path": "/docs/pii-filtering"
     },
     "query-performance": {
       "title": "Query Performance",
--- a/directory.json
+++ b/directory.json
@@ -1000,6 +1000,10 @@
       "title": "Set up traceparent query tag within your application",
       "path": "/docs/opentelemetry/traceparent"
     },
+    "pii-filtering": {
+      "title": "PII Filtering",
+      "path": "/docs/pii-filtering"
+    },
     "query-performance": {
       "title": "Query Performance",
       "path": "/docs/query-performance"
--- a/pii-filtering/index.mdx
+++ b/pii-filtering/index.mdx
@@ -16,11 +16,12 @@ and pg_stat_activity.
 
 In order to strike the right balance between what you share with pganalyze and
 the impact on monitoring, statistics, and insights available through our
-service, we have several collector settings that control metadata collection.
+service, there are several collector settings that control metadata collection.
 
-Note that this filtering happens in the collector, the open-source agent that
-connects to your database and runs on your infrastructure. Filtering occurs
-before the data ever reaches the pganalyze cloud service, or a pganalyze Enterprise Server installation.
+Note that this filtering happens in the collector, the [open-source
+agent](https://github.com/pganalyze/collector/) that connects to your database
+and runs on your infrastructure. Filtering occurs before the data ever reaches
+the pganalyze cloud service, or a pganalyze Enterprise Server installation.
 
 
 ## Filter settings
@@ -41,14 +42,16 @@ The [`filter_query_sample`](/docs/collector/settings#pii-filtering-settings)
 setting determines how content in query samples (individual executions of a
 query collected from logs, which include [Automated EXPLAIN](/docs/explain/setup) plans) is filtered:
  - `all`: do not collect any query samples
- - `normalize`: filter query samples to normalize any literal constants used in the query, and normalize collected EXPLAIN plans to remove constant values
+ - `normalize`: filter query samples and collected EXPLAIN plans to remove any
+   literal constant values and replace them with placeholders
  - `none`: do not filter query samples at all
 
 The [`filter_query_text`](/docs/collector/settings#pii-filtering-settings)
 setting determines how query text from pg_stat_statements is filtered:
  - `none`: do not filter query text at all
  - `unparsable`: filter out queries that could not be parsed by the collector
-Note that these queries are already normalized by pg_stat_statements, so sensitive constants are not a concern here.
+Note that these queries are already normalized by pg_stat_statements, so
+sensitive constants are not a concern here.
 
 
 ## Filtering behavior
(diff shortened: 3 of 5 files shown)


lfittl:
> Maybe this should be broken out to a separate section? It would
> probably also be useful to note what parts of the application are
> impacted, but I left that as a follow-up step (possibly still in this
> patch).

I've been thinking it might be best to have a dedicated docs page even. This is a very common section I link to when discussing the PII filtering / data privacy aspects with customers, and it could benefit from a dedicated page.

Such a page could e.g. have a diagram as well that explains how the filtering happens at the source in the collector, which kind of data gets filtered in a table, etc.

keiko713:
+1 for dedicated page, I think that will be really useful

pii-filtering/index.mdx
```diff
@@ -0,0 +1,73 @@
...
+service, we have several collector settings that control metadata collection.
+
+Note that this filtering happens in the collector, the open-source agent that
+connects to your database and runs on your infrastructure. Filtering occurs
+before the data ever reaches pganalyze.
+
+
+## Filtering settings
+
+The `filter_log_secret` setting determines how content coming from logs is filtered:
```
keiko713, line 25: I wonder if it'd be nice to cross reference this to the above collector setting page. e.g. saying "The `filter_log_secret` setting [in the collector](link to the setting section) determines..."

pii-filtering/index.mdx
```diff
@@ -0,0 +1,73 @@
...
+ - `credential`: Passwords and other credentials, like private keys
+ - `parsing_error`: Text logged by Postgres for parsing errors (could contain anything, including credentials)
+ - `statement_text`: All statement texts (which may contain table data if not using bind parameters)
+ - `statement_parameter`: Bind parameters for a statement (which may contain table data for INSERT statements)
```
keiko713, line 29: it's not only with INSERT statements, I believe (e.g. bind parameter of UPDATEs or DELETEs or even SELECTs can contain some data)

directory.json
```diff
@@ -984,6 +984,10 @@
       "title": "Set up traceparent query tag within your application",
       "path": "/docs/opentelemetry/traceparent"
     },
+    "pii-filtering": {
```
keiko713, line 987: PII filtering only matters if users are using Log Insights, I believe - in that context, I wonder if we can put this under "/docs/log-insights/setup" 🤔
Also in general, this page doesn't mention about Log Insights but it would be nice to mention somewhere to make it clear?

keiko713 (approved):
Have several comments but overall, looking great!

pii-filtering/index.mdx
msakrejda: We can't link to individual settings, but you're right, we should link back to that settings section somewhere.

pii-filtering/index.mdx
msakrejda: I took that from [here](https://github.com/pganalyze/collector/blob/main/state/logs.go#L82), but yeah, we should be clearer here. Maybe
```suggestion
 - `statement_parameter`: Bind parameters for a statement (which may contain table data for statements like INSERT or UPDATE)
```
? Technically you can also get table data in SELECTs and DELETEs, but most of the time, those will just pass an id as a parameter. I think some user discretion is necessary in determining how their particular database fits into this, so we don't need to be exhaustive here, we just need to make it clear the list is not complete.

directory.json
msakrejda: Good point. I'll mention Log Insights and move the page there.

pii-filtering/index.mdx
keiko713: I like the suggested version 👍 agreed that we don't need to be exhaustive here, but former one could wrongly imply that it's only for INSERTs so I think it's nice to update like this

directory.json
msakrejda: Although on second thought... if we capture stuff from, e.g., pg_stat_activity, PII filtering could apply as well, right? I'm not sure that we should put this under Log Insights if it either currently does apply or potentially could apply to data we collect from other sources.

On a related note, though, maybe we should note what sources of data that we capture _may_ contain PII. We do say

>When a third-party tool like pganalyze processes your database
>statistics and logs, it may involve handling data that includes this PII.

but maybe that's too vague? How about I add

>Specifically, pganalyze collects data that may include PII from Postgres logs
>and pg_stat_activity.

lfittl: I'm late to the party, but I disagree with putting this under Log Insights - as you note, some of the settings don't have anything to do with logs, and we usually get this question in a broader context (i.e. which sensitive data is in pganalyze, and how it can be filtered).

Is there a reason we can't put this under the pganalyze collector section?

msakrejda: I put it in its own section since it affects what shows up in the app, but on second thought, so do most of the collector settings :upside_down_face: . I have not moved it under Log Insights due to the above realization. But I think you're right: I'll move it to `collector/pii-filtering.mdx`.

keiko713: Under the collector works for me 👍 I think it'd be nice to put that into ToC rather than being an orphan page

lfittl: I think that was in regards to me taking another look re: wording/how we present things - I can do that tomorrow.

msakrejda: Yeah, I still plan to do that, but those are just mechanical changes. In terms of review, I'm mostly concerned about the accuracy of the Impact section.

pii-filtering/index.mdx
```diff
@@ -0,0 +1,92 @@
...
+In order to strike the right balance between what you share with pganalyze and
+the impact on monitoring, statistics, and insights available through our
+service, we have several collector settings that control metadata collection.
+
+Note that this filtering happens in the collector, the open-source agent that
+connects to your database and runs on your infrastructure. Filtering occurs
+before the data ever reaches pganalyze.
+
+
+## Filtering settings
```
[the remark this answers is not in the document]
lfittl: I also wonder if maybe we should have separate headings named "Filter setting: filter_log_secret", "Filter setting: filter_query_sample" and "Filter setting: filter_query_sample", and then start each section with "**Impact:** ..." and note what we have in the impact section below?

pii-filtering/index.mdx
```diff
@@ -0,0 +1,92 @@
...
+users' privacy and to comply with laws and regulations your organization is
+subject to. When a third-party tool like pganalyze processes your database
+statistics and logs, it may involve handling data that includes this PII.
+
+Specifically, pganalyze collects data that may include PII from Postgres logs
+and pg_stat_activity.
+
+In order to strike the right balance between what you share with pganalyze and
+the impact on monitoring, statistics, and insights available through our
+service, we have several collector settings that control metadata collection.
```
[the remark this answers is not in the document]
msakrejda: I think it's awkward with two "pganalyze"s in the same sentence, but makes sense; I'll reword.

pii-filtering/index.mdx
```diff
@@ -0,0 +1,92 @@
...
+ - `ops`: System errors, network errors, file locations, pg_hba.conf contents, and configured commands (e.g. archive command)
+ - `unidentified`: Text that could not be identified and might contain secrets
+
+And of course, `all` and `none` apply all and none of the above filters, respectively.
+
+The [`filter_query_sample`](/docs/collector/settings#pii-filtering-settings)
+setting determines how content in query samples (individual executions of a
+query collected from logs) is filtered:
+ - `all`: do not collect any query samples
+ - `normalize`: filter query samples to normalize any literal constants used in the query
```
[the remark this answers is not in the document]
msakrejda: Good call on including EXPLAIN plans, but I'll reword this slightly to use a parallel construction, since we're doing something very similar in both cases.

collector/pii-filtering.mdx
```diff
@@ -0,0 +1,97 @@
...
+### Query samples / Automated EXPLAIN
+
+**Setting**: `filter_query_sample` / `FILTER_QUERY_SAMPLE`
+
+This setting determines how content in query samples (individual executions of a
+query collected from logs, which include [Automated EXPLAIN](/docs/explain/setup) plans) is filtered:
+ - `all`: do not collect any query samples
+ - `normalize`: filter query samples and collected EXPLAIN plans to remove any
+   literal constant values and replace them with placeholders
+ - `none`: do not filter query samples at all
```
msakrejda, line 74: @lfittl I noticed we only really talk about EXPLAIN plans with the `normalize` setting. Should we add that to the others?

lfittl: Yeah, maybe good for consistency, I don't think readers make that link automatically.

PR #309, opened 2025-02-19
Fix query tuning helper

msakrejda:
The backslashes in the regexp we use must be escaped for JavaScript.


lfittl (approved):
Ah, good catch!

PR #310, opened 2025-02-27
Fix external links in embedded log insights page

msepga:
When embedded in the app, these links previously redirected to the wrong domain (https://app.pganalyze.com/docs/...). Here, we fix the link target.


keiko713 (approved):
Ah interesting, I guess this won't work because these two pages are not loaded in the app. Thanks for the fix!

PR #315, opened 2025-03-14
Remove Cosmos DB from the title, add note about Cosmos DB

keiko713:
As discussed internally, remove this from the officially supported list. I still kept some special tweaks note for Cosmos DB in the installation guide docs, but I added some caution in the beginning of the guide.

### DIFF
--- a/install/azure_database/01_configure_azure_instance.mdx
+++ b/install/azure_database/01_configure_azure_instance.mdx
@@ -17,7 +17,7 @@ extension for collecting query statistics.
 This guide assumes you have an already running Azure Database for PostgreSQL flexible server you want to monitor.
 
 For Azure Cosmos DB for PostgreSQL Cluster, this step is not required as pg_stat_statements is already preloaded.
-While the pganalyze collector supports collecting data from Azure Cosmos DB for PostgreSQL, this integration is not officially supported at this time.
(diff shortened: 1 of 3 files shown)


install/azure_database/01_configure_azure_instance.mdx
```diff
@@ -15,7 +15,10 @@ to first follow these steps to enable the [pg_stat_statements](http://www.postgr
 extension for collecting query statistics.
 
 This guide assumes you have an already running Azure Database for PostgreSQL flexible server you want to monitor.
+
 For Azure Cosmos DB for PostgreSQL Cluster, this step is not required as pg_stat_statements is already preloaded.
+While the pganalyze collector supports collecting data from Azure Cosmos DB for PostgreSQL, this integration is not officially supported at this time.
+This means that while data collection may work, we cannot guarantee full compatibility, and certain features may not function as expected.
```
msakrejda, line 21: I think the current wording implies that it might be supported again at some point in the future, which as I understand it is not the case. How about something like

```suggestion
While the pganalyze collector supports collecting data from Azure Cosmos DB for PostgreSQL, this integration is no longer officially supported.
This means that while data collection may work, we cannot guarantee full compatibility, and certain features may not function as expected.
```

?

Ideally, we'd also be able to point to a deprecation notice from MS to head off questions about why, but a quick search did not turn up anything like that, so it is what it is.

msakrejda (approved):

PR #319, opened 2025-03-22
Add Important Fields to Function Scan node

msakrejda:
Missed in aa025a5e6006b944699046d09a37105d7807b067. That commit added
some fields to the list, but did not add them to the frontmatter
metadata that is used to show this field data in some EXPLAIN views.


keiko713 (approved):
For reference: https://github.com/pganalyze/pganalyze-docs/commit/aa025a5e6006b944699046d09a37105d7807b067#diff-c046998bf0cd8bd4b684d39f1973712406643b828468f0d6b386864faa0d0d0aR16

PR #320, opened 2025-03-24
Polish collector settings docs

msakrejda:
- Clarify automatic collector reload behavior
- Drop commented-out PGA_ENABLE_REPORTS section
- Collector settings cleanup (see details in commit message)

### DIFF
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -25,8 +25,9 @@ name the other sections after the servers they correspond to, though note these 
 in the app. In-app names are based on hostname settings as determined during monitoring.
 
 After you make changes, you can run `pganalyze-collector --test` to verify the new configuration.
-If the test succeeds, this will automatically load the new configuration in the collector background process
-(equivalent to using the `--reload` flag). This minimizes monitoring interruptions and simplifies config file updates.
+For package-based installs, if the test succeeds, this will automatically load the new configuration in the collector
+background process (equivalent to using the `--reload` flag). This minimizes monitoring interruptions and simplifies
+config file updates.
 
 The tables below list configuration settings, their defaults if not set, and their descriptions. If a setting is
 configurable through environment variables, the environment variable name follows the setting in parentheses.
--- a/collector/settings.mdx
+++ b/collector/settings.mdx
@@ -70,14 +70,6 @@ Common settings for configuring collector behavior, independent of the platform.
         Note: The name of the environment variable was <code>PGA_API_BASEURL</code> prior to collector release v0.55.0.
       </td>
     </tr>
-    {/*
-      omit this for now
-      <tr>
-        <td>enable_reports (<code>PGA_ENABLE_REPORTS</code>)</td>
-        <td>false</td>
-        <td></td>
-      </tr>
-    */}
     <tr>
       <td>skip_if_replica (<code>SKIP_IF_REPLICA</code>)</td>
       <td>false</td>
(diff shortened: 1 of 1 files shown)


collector/settings.mdx
```diff
@@ -133,7 +126,7 @@ helper functions are expected to be defined, and the one we connect to for serve
     </tr>
     <tr>
       <td>db_password (<code>DB_PASSWORD</code>)</td>
-      <td>[none]</td>
+      <td>n/a, either this or db_url is required</td>
```
lfittl, line 129: I'm not sure that its correct that a password is required - AFAIK we can successfully connect over a unix domain socket with peer authentication if that's set up (or a TCP-based connection with trust, e.g. when restricting to local connections).

collector/settings.mdx
```diff
@@ -409,13 +402,12 @@ Only relevant if you are running your database in Azure using Azure Database for
     <tr>
       <td>azure_ad_tenant_id (<code>AZURE_AD_TENANT_ID</code>)</td>
       <td>[none]</td>
-      <td>The "Directory (tenant) ID" on your application. Only necessary if not using the recommended Managed Identity setup; see these <a href="https://pganalyze.com/docs/log-insights/setup/azure-database#setting-up-authentication-using-azure-ad-application-alternative-to-managed-identity">setup instructions</a>
-      for details</td>
+      <td>The "Directory (tenant) ID" on your application. Only necessary if not using the recommended Managed Identity setup; see these <a href="https://pganalyze.com/docs/log-insights/setup/azure-database#setting-up-authentication-using-azure-ad-application-alternative-to-managed-identity">setup instructions</a> for details</td>
```
lfittl, line 405: Not new in this PR, but it feels slightly better to use "the" instead of "these" here to me:
```suggestion
      <td>The "Directory (tenant) ID" on your application. Only necessary if not using the recommended Managed Identity setup; see the <a href="https://pganalyze.com/docs/log-insights/setup/azure-database#setting-up-authentication-using-azure-ad-application-alternative-to-managed-identity">setup instructions</a> for details</td>
```

collector/settings.mdx
```diff
@@ -409,13 +402,12 @@ Only relevant if you are running your database in Azure using Azure Database for
...
     </tr>
     <tr>
       <td>azure_ad_client_id (<code>AZURE_AD_CLIENT_ID</code>)</td>
       <td>[none]</td>
-      <td>The "Application (client) ID" on your application. Alternatively, when using a Managed Identity, this can also be used to define which identity to use, when a VM has multiple identities assigned.</td>
+      <td>The "Application (client) ID" on your application. Only necessary if not using the recommended Managed Identity setup (see above). Alternatively, when using a Managed Identity, this can also be used to define which identity to use, when a VM has multiple identities assigned.</td>
```
lfittl, line 410: Is the "," at the end correct? i.e. shouldn't it be
```suggestion
      <td>The "Application (client) ID" on your application. Only necessary if not using the recommended Managed Identity setup (see above). Alternatively, when using a Managed Identity, this can also be used to define which identity to use when a VM has multiple identities assigned.</td>
```

collector/settings.mdx
msakrejda: It's not, but it's required in typical situations. I felt that this would be clearer. But I don't feel strongly about it; I'll drop this change.

lfittl: Yeah, makes sense - I do think its better to document how the code behaves here (i.e. it doesn't require it), and if needed we could add additional commentary as to what's typical (but I think its okay if we omit here, since I think most people understand they'd be using a password).

keiko713 (approved):
Thanks for the update!

collector/settings.mdx
msakrejda, line 129:
```suggestion
      <td>[none]</td>
```

PR #322, opened 2025-04-08
Add grant statement in Log Insights docs too

keiko713:
Got feedback that it would be nice to restate in this section too. As it won't hurt adding it, I simply added the line.

### DIFF
--- a/directory.json
+++ b/directory.json
@@ -2248,7 +2248,10 @@
     "Function Scan": {
       "shortDescription": "Scans the result of a set-returning function (like unnest or regexp_split_to_table).",
       "importantFields": [
-        "Function Name"
+        "Function Name",
+        "Function Call",
+        "Filter",
+        "Rows Removed by Filter"
       ],
       "path": "/docs/explain/scan-nodes/function-scan"
     },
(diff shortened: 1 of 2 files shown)


log-insights/setup/crunchy-bridge/01_create_helper_function_and_test.mdx
```diff
@@ -17,6 +17,10 @@ Log output for Crunchy Bridge database servers is collected directly from the da
 
 If you are using the application user, an additional function needs to be created to grant access to the logs (**run the following with superuser credentials**):
 
+```sql
+GRANT pg_monitor TO application;
+```
```
msakrejda, line 22: This will end up in a separate code block, right? That might look a little weird. Maybe we should be more explicit:

````suggestion
If you are using the application user, first, **connect with superuser credentials** and ensure this user has the `pg_monitor` role:
```sql
GRANT pg_monitor TO application;
```
Then, create a function to grant this user access to the logs:
````

**Review by msakrejda — APPROVED** [review]

File: log-insights/setup/crunchy-bridge/01_create_helper_function_and_test.mdx (hunk 1 above)
Comment (keiko713) [reply log-insights/setup/crunchy-bridge/01_create_helper_function_and_test.mdx:22]: Ah yeah, I got lazy, looking better 👍

File: directory.json [hunk 2]
```diff
@@ -2248,7 +2248,10 @@
     "Function Scan": {
       "shortDescription": "Scans the result of a set-returning function (like unnest or regexp_split_to_table).",
       "importantFields": [
-        "Function Name"
+        "Function Name",
+        "Function Call",
+        "Filter",
+        "Rows Removed by Filter"
```
keiko713, line 2254: Looks like this change was missed in https://github.com/pganalyze/pganalyze-docs/pull/319
(need to run gatsby locally to re-generate this file)

PR #324, opened 2025-05-07
Adding config variable for cloudsql

ryanbooz:


collector/settings.mdx
```diff
@@ -485,6 +485,12 @@ Only relevant if you are running your database in GCP using Google Cloud SQL or
       <td>n/a, required for IAM Authentication</td>
       <td>GCP region of the Cloud SQL instance</td>
     </tr>
+    <tr>
+      <td>gcp_use_public_ip (<code>GCP_USE_PUBLIC_IP</code>)</td>
+      <td>false</td>
+      <td>Set to 'true' if the CloudSQL proxy connextion should use the available public IP rather than the default private IP. Setting
+        to true when no public IP is assigned will result in a connection error.</td>
```
msakrejda, line 492: Two minor things: typo on `connection` and I think we usually wrap config setting values referenced in the description in `code` tags:
```suggestion
      <td>Set to <code>true</code> if the CloudSQL proxy connection should use the available public IP rather than the default private IP. Setting
        to <code>true</code> when no public IP is assigned will result in a connection error.</td>
```
Looks good otherwise; thanks for updating!

msakrejda (approved):

PR #325, opened 2025-05-19
Retitle Log Insights events

msakrejda:
Use terser titles for Log Insights log classifications. This will let
us update the UI to make log classifications easier to recognize.


lfittl (approved):
:+1: Thanks for making these updates!

PR #329, opened 2025-06-17
Add normalized query to getQueryStats documentation

seanlinsley:
`getQueryStats` supports returning the full normalized query, but that wasn't in the public documentation. This change was prompted by a support request asking for a way to export the full query text.


keiko713 (approved):
It's actually interesting that it wasn't documented before... did we have any reason for that? Looks good otherwise.

seanlinsley:
Customers should have access to the full query text from the API since it's their data, so I assume this was just missed when writing the documentation.

PR #337, opened 2025-07-15
Add EXPLAIN Grid docs

msakrejda:


explain/grid.mdx
```diff
@@ -0,0 +1,82 @@
...
+## Metrics
+
+The execution metric to show can be selected at the top of the plan:
+
+  - **Est. Cost**: The cost of the node according to the Postgres planning model,
+    excluding the cost of the node's children.
+  - **Runtime**: How much time was spent executing this node, excluding time spent
+    executing the node's children.
+  - **Rows**: How many rows each node produced, and how many it was expected to
+    produce according to the query plan
```
keiko713, line 44: Looks like some lines have periods at the end and some don't, would be good to unify

explain/grid.mdx
```diff
@@ -0,0 +1,82 @@
...
+  - **Buffers**: How many buffers were hit in the buffer cache and read from disk
+    (these are combined) or dirtied while executing this query, excluding
+    buffers touched while executing the node's children
+  - **Reads**: How much data was read from disk (or the OS page cache) and how much
```
keiko713, line 48: Likely not a question to ask here but I was curious - when track_io_timing is off, how does this look like?

keiko713 (approved):
Tiny comments but looks good!

explain/grid.mdx
msakrejda: Ah, I tried to, but I guess I missed some. Thanks!

explain/grid.mdx
msakrejda: It should be `-`, but I think we actually show zeros; I'll fix that separately. We should also maybe show why: it's not that hard for us to check the setting.

PR #352, opened 2025-08-12
Rename query tuning advisor check name

keiko713:
In passing:

* specify the default config value of queries/advisor_insight check (currently causing the web build failure)
* add a check to TOC
* regen directory.json

### DIFF
--- a/directory.json
+++ b/directory.json
@@ -308,6 +308,10 @@
       "title": "2025.03.3 Release",
       "path": "/docs/enterprise/releases/2025-03-3"
     },
+    "enterprise/releases/2025-03-4": {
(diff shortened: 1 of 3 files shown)


directory.json
```diff
@@ -308,6 +308,10 @@
       "title": "2025.03.3 Release",
       "path": "/docs/enterprise/releases/2025-03-3"
     },
+    "enterprise/releases/2025-03-4": {
+      "title": "2025.03.4 Release",
+      "path": "/docs/enterprise/releases/2025-03-4"
+    },
```
lfittl, line 314: Ah, good catch - my bad.

PR #357, opened 2025-08-20
Add a check documentation for AdvisorInsight

keiko713:


components/CheckDocumentation/queries/AdvisorInsight.tsx
```diff
@@ -0,0 +1,86 @@
...
+> = () => {
+  return (
+    <div>
+      <p>
+        Some common patterns that lead to suboptimal performance are found for
+        this query. You can check out the exact pattern (insight) in the
+        insights section of "Query Summary" panel. Tweaking this query may
+        improve the performance of this query.
+      </p>
+      <h4>Next Step</h4>
```
lfittl, line 16:
```suggestion
      <h4>Next Steps</h4>
```

components/CheckDocumentation/queries/AdvisorInsight.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <p>
+        You can tune a query using Workbooks.
```
lfittl, line 18:
```suggestion
        You can test a Query Advisor insight using Workbooks.
```

components/CheckDocumentation/queries/AdvisorInsight.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <ul>
+          <li>
+            <h5>Create a new workbook and record the baseline</h5>
+            <p>
+              For each insight, there is a "Create Workbook" button to create a
+              new workbook using this query. Create a new workbook using this
+              button, then follow the steps to record the baseline. This is also
+              important step to make sure that the insight can be found with the
+              baseline too. If the query has multiple query plans, it's
```
lfittl, line 27:
```suggestion
              ```suggestion
              button, then follow the steps to record the baseline. This step
              re-confirms that the insight still reproduces in the baseline too.
              If the query has multiple query plans, it's
```

components/CheckDocumentation/queries/AdvisorInsight.tsx
```diff
@@ -0,0 +1,86 @@
...
+              to create a new variant. Each insight usually contains some
+              remediation suggestion, usually by a query rewrite. You can apply
+              this rewrite to the baseline query (current query), then record
+              EXPLAIN plans of it. With this, you can check whether the rewrite
+              improved performance and/or removed the existing insight.
+            </p>
+          </li>
+        </ul>
+      </p>
+      <h4>Common Patterns</h4>
```
lfittl, line 45: I would omit this, since its duplicative of the insights themselves (except for GROUP BY, but we're going to add that soon anyway). We can expand on the kinds of insights in the Query Advisor documentation which we could link from here once its added.

components/CheckDocumentation/queries/AdvisorInsight.tsx
```diff
@@ -0,0 +1,86 @@
...
+        the Query Optimization Examples documentation
+      </a>
+      .
+    </div>
+  );
+};
+
+const documentation: CheckDocs = {
+  description:
+    "Checks EXPLAIN Plans captured by Automated EXPLAIN and creates insights for any query tuning opportunities are found.",
```
lfittl, line 81:
```suggestion
    "Checks EXPLAIN Plans captured by Automated EXPLAIN and creates Query Advisor insights for tuning opportunities.",
```

components/CheckDocumentation/queries/AdvisorInsight.tsx
keiko713: > This step re-confirms that the insight still reproduces in the baseline too.

I think this is grammatically a bit incorrect, should be "still can be reproduced" if we wanna go this way.

components/CheckDocumentation/queries/AdvisorInsight.tsx
keiko713: Uhmm, this doesn't take much space, so I feel like it's okay to have even if it's duplicative, but what do you think? 🤔 

image

Also, the current inefficient nested loop insight doesn't have the information about the insight that much actually. ORDER BY does have more though. We can certainly have more info below instead of above, that's also alternative.

image
image

> We can expand on the kinds of insights in the Query Advisor documentation which we could link from here once its added.

I agree that having the link to the docs page will be helpful and we sure need to have more detailed info over there, so I was more focusing on having one-liners here.

PR #359, opened 2025-08-25
Broken link in enterprise documentation reported by user

ryanbooz:
Also updated CSS for the code block area in the main pganalyze-web repo for improving right padding to avoid copy button overlapping with code content.

### DIFF
--- a/enterprise/setup/index.mdx
+++ b/enterprise/setup/index.mdx
@@ -47,6 +47,7 @@ You should now be able to pull the pganalyze Enterprise Server Docker image:
 <CodeBlock language="bash">
 {`docker pull quay.io/pganalyze/enterprise:[version]`}
 </CodeBlock>
+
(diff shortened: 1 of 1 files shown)


seanlinsley:
It looks like this PR is missing the described changes: the only included change is an extra newline

ryanbooz:
> It looks like this PR is missing the described changes: the only included change is an extra newline

This is a rendering issue with the component that we couldn't easily identify in earlier testing. Something about how the block gets rendered without a break from the previous line messes up the link render. There's a chance that updating the MDX processor might fix some of these edge cases, but in a previous attempt by both myself and @msakrejda, doing so was non-trivial and we decided some of these quirks were OK to handle for right now.

Please see the updated staging deployment for verification in https://github.com/pganalyze/pganalyze-web/pull/699

PR #361, opened 2025-09-24
Docs for Query Advisor

jensy:


query-advisor/query-advisor-insight-issue.png
keiko713, line 1: Gatsby is doing the thingy again with this image 😢 

Image

query-advisor/from-insight-to-conclusion.mdx
```diff
@@ -0,0 +1,199 @@
...
+2. **Apply insights** - Use suggested rewrites to create variants
+3. **Benchmark results** - Compare performance improvements systematically
+4. **Validate changes** - Ensure optimizations work across parameter sets
+5. **Apply to production** - Implement verified improvements with confidence
+
+## Step 4: Apply Query Advisor insight
+
+In the workbook variant editor, Query Advisor provided an **"Apply"** button for the detected insight:
+
+![Workbooks interface showing Query Advisor sidebar with apply insight functionality](workbooks-query-advisor-sidebar.png)
```
keiko713, line 94: This screenshot might be wrong? It's showing the variant detail page, and it doesn't have the apply insight button.

query-advisor/from-insight-to-conclusion.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+### Automatic rewrite
+Query Advisor automatically modified the query to force better index selection:
+
+<CodeBlock language="sql">
+{`SELECT databases.*
+FROM databases
+WHERE databases.server_id = $1
+  AND databases.hidden = $2
+ORDER BY databases.id + $zero ASC  -- Added +0 to prevent index sort usage
```
keiko713, line 106: I think it's a bit confusing to see the normal param refs ($1, $2) and pganalyze named param refs ($zero). Might be good to use what we can see in the screenshot (pganalyze param refs, like $database_id) for the consistency?

query-advisor/from-insight-to-conclusion.mdx
```diff
@@ -0,0 +1,199 @@
...
+## Step 6: Validate the results
+
+The dramatic improvement confirmed Query Advisor's analysis was correct:
+
+### Why Postgres chose poorly
+Postgres's cost-based planner estimated it could find the limited rows quickly by scanning the primary key index in sort order. This failed because filtering happened after sorting, requiring examination of many irrelevant rows.
+
+### Why the rewrite works
+By preventing sort-order optimization, Postgres had to:
+1. Use the index_databases_on_server_id_and_datname index for efficient filtering
```
keiko713, line 147: As you were saying "Primary key scan → server_id + datname index" previously in line 134, it might read better to use that one too (e.g. "Use the server_id + datname index for efficient filtering")

query-advisor/from-insight-to-conclusion.mdx
```diff
@@ -0,0 +1,199 @@
...
+- **Plan A (Baseline)**: Gather Merge → Index Scan on databases_pkey
+- **Plan B (Variant 1)**: Limit → Sort → Bitmap Heap Scan → Bitmap Index Scan
+- **Runtime improvement**: 495.51ms → 58.11ms
+- **Index strategy**: Primary key scan → server_id + datname index
+
+![Workbooks plan comparison showing performance improvements between baseline and optimized query variants](workbooks-compare-plans.png)
+
+## Step 6: Validate the results
+
+The dramatic improvement confirmed Query Advisor's analysis was correct:
```
keiko713, line 140: Uhm, this sentence is strange - or maybe sentence itself is okay, but it can end with "." instead of ":"

query-advisor/where-to-find.mdx
```diff
@@ -0,0 +1,60 @@
+---
+title: 'Where to Find Query Advisor'
+backlink_href: /docs/query-advisor
+backlink_title: 'Query Advisor'
+---
+
+Once Query Advisor is configured and collecting EXPLAIN plans, optimization insights appear in several locations throughout pganalyze:
```
keiko713, line 7: (feel free to disregard) I don't have good alternatives, but this sounds like there is some special configuration is needed for Query Advisor. In Getting Started, it says "Query Advisor requires the pganalyze collector to be properly configured for query monitoring", which is more accurate - although I'm not sure putting all of these sentences here makes sense 🤔

query-advisor/where-to-find.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+## Query Advisor Landing Page
+
+Access the dedicated Query Advisor page to view all detected optimization opportunities across your databases, organized by potential impact.
+
+![Query Advisor landing page showing automated EXPLAIN plan analysis with detected optimization opportunities](query-advisor-landing.png)
+
+As pganalyze collects EXPLAIN plans through auto_explain, Query Advisor automatically analyzes them and flags problematic patterns in near real-time.
```
keiko713, line 15: I think we usually say near real-time for the checks that are running per compact snapshots (aka every 10 seconds), but this check runs daily.

Image

query-advisor/query-detail-page.png
keiko713, line 1: It's okay as is, but Bison pushed the change to update the "Tune query in workbook" link to "Create Workbook", so you could potentially update this screenshot.

query-advisor/alerts.mdx
```diff
@@ -0,0 +1,34 @@
...
+title: 'Track Query Advisor Insights With Alerts'
+backlink_href: /docs/query-advisor
+backlink_title: 'Query Advisor'
+---
+
+Query Advisor automatically detects optimization opportunities, but you need to configure alerts to get notified when new insights are discovered. This guide covers setting up alerts for Query Advisor insights and integrating with your notification workflows.
+
+## Alert configuration overview
+
+Query Advisor checks are **enabled by default** for all databases in your pganalyze organization, but **alerts must be explicitly configured** through alert policies to receive notifications about newly detected optimization opportunities.
```
keiko713, line 11: Should we mention about plans? Technically it's not enabled by default for all databases, and we actually have quite a lot of customers with legacy plans.

query-advisor/alerts.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+### Default behavior
+- **Continuous monitoring**: Query Advisor analyzes all incoming EXPLAIN plans automatically
+- **Background detection**: Insights are generated and stored without notifications
+- **Dashboard visibility**: All insights appear on the Query Advisor landing page
+- **No notifications**: You won't receive alerts until explicitly configured
```
keiko713, line 17: Any new customers, this is correct, but we will enable alerts for some existing customers I believe. Not sure how to mention that in the docs though.

query-advisor/insights/index.mdx
```diff
@@ -0,0 +1,91 @@
...
+### Query rewrites
+- **Materialized CTEs**: Forces Postgres to materialize intermediate results to prevent inefficient nested loops
+- **Sort prevention**: Uses arithmetic expressions (e.g., `+0`) to prevent Postgres from using index sort order
+
+## Detection methodology
+
+Query Advisor analyzes EXPLAIN plans through a **two-phase process**:
+
+### Phase 1: Initial screening
+- Processes **7 million+ plan samples daily** in real-time
```
keiko713, line 47: I'm not sure if this number is needed. Something like this can be outdated easily, also the customer may care about how their query is analyzed, but maybe don't care much about how the pganalyze app analyzes other customers' query, especially in this context? I think this is here to lead to the "scalable" part in the line 57... maybe something like this might fit better in the blog post than the documentation?

keiko713 (commented):
Just did the first pass, I think it's covering all the things that we need to mention, so from that point of view, it's great 🎉 Thanks for putting this together!!
Overall, I found it pretty... marketing heavy(?), really "selling" about this feature. I understand that some aspect of that is needed, to provide the confidence on the product/feature, so I'm not strongly against it, I just wanted to note that I noticed it.
I also noticed some parts might be better fit in the "behind the scenes" type of blog post than the docs (left some comment about it too), but I don't think we have much time to re-structure and the overall structure looks good so maybe it's fine to leave as is.

query-advisor/getting-started.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+Query Advisor currently detects and provides solutions for these common Postgres performance issues:
+
+### Inefficient Nested Loops
+Detects when Postgres chooses nested loop joins due to row estimation mismatches, suggesting hash or merge join alternatives through query rewrites.
+
+### Wrong Index Due To ORDER BY
+Identifies queries where Postgres uses an index only for sorting while filtering large numbers of rows, suggesting modifications to force better index utilization.
+
+More optimization patterns are continuously being added. See the complete list in [Supported Insights](/docs/query-advisor/insights).
```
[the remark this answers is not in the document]
jensy: Dunno if we want to document things that are not real.

query-advisor/alerts.mdx
jensy: I think we should document what the default behavior is. But I could add a sentence along the lines of "If Missing Index alerts are enabled, these will be enabled by default, too" @lfittl ?

query-advisor/query-advisor-insight-issue.png
jensy: I only have issues on my local deploy, the preview deploy looks fine. But yeah, not really sure what to do about this…

query-advisor/where-to-find.mdx
lfittl: To confirm, for now I think its better to call it daily, since that's what happens to be visible to the user. The fact that we do near-realtime analysis as a first phase is probably not worth sharing in the docs.

jensy:
Pull "Where to find QA" page into "Getting Started" page.

query-advisor/alerts.mdx
jensy: Mention "For new customers, these will not be enabled by default"

Also, remove mention of policies, include Slack.

PR #392, opened 2026-03-28
Migrate to Astro

msakrejda:
This updates the docs mostly for MDXv3 compatibility, and a few other
changes to better work in Astro instead of Gatsby.


keiko713 (approved):
Nicely addressed!

PR #394, opened 2026-04-13
Remove aurora_stat_plans recommendation

seanlinsley:
A customer reported that despite increasing `pg_stat_statements.max` to 50,000, deallocations occurred around every 30 minutes, causing significant `LWLock:pg_stat_statements` lock contention. They were able to resolve the issue by setting `aurora_compute_plan_id = 0`, which disables aurora_stat_plans.

Until we have a better understanding of the issue, this PR removes the recommendation to enable aurora_stat_plans during collector setup.


lfittl (approved):
Thanks - as discussed we've had a few reports of this issue, so removing it for now seems best, until AWS has addressed this.

PR #395, opened 2026-04-13
Initial draft for install and upgrade reorganization and additions

ryanbooz:
In multiple recent Enterprise installs and upgrades, customers and DevOps teams were looking for a more coherent overview of how the pieces fit together. Often, multiple teams are responsible for each piece of the installation.

This PR is an attempt to provide that overview for both initial installation and upgrade. After laying out the architecture and process, individual guides (pre-existing) dive into the process for each.

I've also added upgrade documentation for AKS (EKS to follow soon) and reorganized the Enterprise tree just slightly.

I understand this could feel like overkill, but my experience with customers has shown that what feels straightforward to us isn't always to the teams managing the installation, especially when multiple platforms and teams are in play. I think the overview pages can certainly be tightened up and improved, but the overall "here's the road ahead" take has been helpful for multiple users to get started. 

Of course, open to any and all discussion to make these as helpful as possible!


keiko713:
I like the direction! I think it's a good improvement, and I don't think this is an overkill.
I wonder if we could somehow add architecture diagram to Setup Overview. I usually find it useful to see things in the diagram when trying to understand a new thing. However, it might be too much info, as Setup Overview already contains lots of info.

ryanbooz:
> I like the direction! I think it's a good improvement, and I don't think this is an overkill. I wonder if we could somehow add architecture diagram to Setup Overview. I usually find it useful to see things in the diagram when trying to understand a new thing. However, it might be too much info, as Setup Overview already contains lots of info.

Thanks for the feedback and support. I had a similar thought with the diagram and think it could be worked in somehow. As I said in the previous PR with that diagram, I think it's a good high-level concept and could fit well here. Or, at the very least, we could find a good place in here to link to it. Should that architecture diagram page maybe move up in the tree on the left?

enterprise/setup/docker.mdx
```diff
@@ -0,0 +1,211 @@
...
+The pganalyze container image is mostly self-contained, and runs various services inside the Docker (or podman) container.
+
+<ToC items={props.toc} />
+
+## Installation steps
+
+### Pre-requisites
+
+* Provision a virtual machine
+  * Assign at least 4GB - 8GB of RAM, 1-2 dedicated CPU cores
```
msakrejda, line 21: >...at least 4GB - 8GB...

As a reader, I find this kind of language confusing: is the minimum 4 or 8? How do I reason about sizing?

msakrejda: Oh I see, this is just moved. We can leave it.

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
+- **Statistics database** — a PostgreSQL database that you provision and manage. The
+  pganalyze Enterprise Server uses it to store all collected metrics, query statistics, and
+  configuration data. This is separate from the databases you are monitoring.
 
-The Docker image is made available to you through the container hosting service quay.io.
+- **pganalyze collector** — a lightweight agent that connects to your monitored PostgreSQL
+  servers, gathers performance data, and reports it back to the pganalyze Enterprise Server.
+  The collector can run as part of the pganalyze Enterprise Server container (simplest for
+  getting started) or as a [separate installation](/docs/enterprise/setup/separate-collector-install)
+  (recommended for production deployments with multiple servers or network boundaries).
```
msakrejda, line 36: This is a little different than the advice we give [here](https://github.com/pganalyze/pganalyze-docs/pull/395/changes#diff-43ea0f6966ca4b68b89c5be400a2ca016732b7690aecd459fb60e2937d8af540R175-R176):

> 1. Use the integrated collector in the pganalyze Enterprise Server image (works best for Amazon RDS / Aurora)
>2. Use a [separate collector installation](/docs/enterprise/setup/separate-collector-install) (recommended all other environments)

We should be consistent.

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
-Create the initial user:
-
-<CodeBlock language="bash">
-{`docker run --env-file /etc/pganalyze.env quay.io/pganalyze/enterprise:[version] \
-  rake db:seed`}
-</CodeBlock>
+---
 
-Make sure to note down the default user email and password:
+### Phase 1: Deploy the pganalyze Enterprise Server
```
msakrejda, line 51:
```suggestion
### Phase 1: Deploy pganalyze Enterprise Server
```

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
+[firewall configuration guide](/docs/enterprise/setup/firewall-configuration) for details.
 
-Before you can add a database to the pganalyze installation, you'll need to enable the `pg_stat_statements` extension on it.
+**4. Run initial setup and verify**
 
-This extension is bundled with PostgreSQL and allows you detailed query monitoring. Its overhead on the database system is minimal, and it is required to use pganalyze.
+Each deployment guide includes steps to initialize the database schema, run the pganalyze
+Enterprise Server self-check, and create an initial admin user. Once complete, you can log
+in to the pganalyze web UI, create your organization, and optionally get the required API key(s) to
+allow an separately installed collector to submit monitoring data.
```
msakrejda, line 104:
```suggestion
allow a separately installed collector to submit monitoring data.
```

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
+   plan used at query time, including buffer usage and timing data.
 
-![](server_empty.png)
+#### For each monitored server
 
-You can check whether any information has been received by clicking the "Server Settings" link in the left navigation, and scrolling down to the "Debug Info" section:
+- **Enable `pg_stat_statements`** — this extension is bundled with PostgreSQL but not
+  enabled by default in most environments. It must be added to `shared_preload_libraries`, 
+  which requires a database restart. If you plan to also enable `auto_explain`, add both at the 
+  same time to avoid a second restart.
```
msakrejda, line 141: Is this still true? Last time I checked (admittedly a while ago, so I may be misremembering), I believe RDS includes pg_stat_statements (I think it does not create the extension for you, but it was in `shared_preload_libraries` by default already).

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
 
-![](debug_info_no_data.png)
+- **Create a monitoring user and helper functions** — we recommend a dedicated user with the
+  `pg_monitor` role, which provides read access to all the statistics views pganalyze needs
+  without requiring superuser privileges.
 
-Once data is coming in successfully, the "Debug Info" looks like this:
+- **Configure log delivery** (for Log Insights) — the setup varies by platform. See
+  your platform's installation guide for specific instructions.
```
msakrejda, line 148: Should we link to https://pganalyze.com/docs/log-insights/setup here?

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
 
-<CodeBlock language="bash">
-{`docker stop pganalyze
-docker rm pganalyze
-docker run -d --name pganalyze --env-file /etc/pganalyze.env -p 8080:5000 \
-  quay.io/pganalyze/enterprise:[version]`}
-</CodeBlock>
+- [Amazon RDS and Aurora](/docs/install/amazon_rds/00_overview)
+- [Azure Database for PostgreSQL](/docs/install/azure_database/01_configure_azure_instance)
+- [Self-managed PostgreSQL servers](/docs/install/self_managed/00_choose_setup_method)
```
msakrejda, line 178: I don't think Guided Setup works for Enterprise Server, does it? We should maybe just link to

```suggestion
- [Self-managed PostgreSQL servers](/docs/install/self_managed/01_create_monitoring_user)
```

enterprise/upgrade/docker.mdx
```diff
@@ -0,0 +1,99 @@
...
+If this returns any errors, review your environment configuration in `/etc/pganalyze.env`.
+
+### Step 6: Start the container
+
+<CodeBlock language="bash">
+{`docker run -d --name pganalyze -p 8080:5000 --env-file /etc/pganalyze.env \
+  quay.io/pganalyze/enterprise:[new version]`}
+</CodeBlock>
+
+At this point the upgrade is complete and the pganalyze UI should be accessible as expected.
```
msakrejda, line 76:
```suggestion
At this point the upgrade is complete and the pganalyze UI should be accessible.
```

enterprise/upgrade/index.mdx
```diff
@@ -0,0 +1,102 @@
...
+DropTableStatsPacked.new.up`}
+</CodeBlock>
+
+Until the migration is completed the pganalyze UI will only show new data. After the
+migration is complete the full history will be available again.
+
+### Upgrading to 2018.08 and newer: license key requirement
+
+Starting in pganalyze Enterprise Server
+[2018.08.0](/docs/enterprise/releases/2018-08-0), the pganalyze container requires the
```
msakrejda, line 91: Do we have anyone using releases older than this? Not a blocker (I'm assuming this is moving existing copy), but we can probably remove this in a follow-up.

enterprise/upgrade/index.mdx
```diff
@@ -0,0 +1,102 @@
...
+notes that apply across all deployment environments. For environment-specific steps, see
+the guides linked below.
+
+<ToC items={props.toc} />
+
+## General upgrade process
+
+Regardless of how you deploy the pganalyze Enterprise Server, every upgrade follows the
+same sequence:
+
```
msakrejda, line 21: This is maybe unnecessary, but I think it's good practice when upgrading in general:

```suggestion
0. ** Review release notes** — check the [release notes](https://pganalyze.com/docs/enterprise/releases) for the new version (and any intermediate versions you are upgrading past). Releases with extra migration steps are noted in the following section, but release notes will help you know what to expect.

```

msakrejda (approved):
I left some comments, but overall this looks good. We're duplicating here more than we've done in the past (where we've used shared mdx fragments), but I think that's okay, since trying to avoid duplication can get complicated.

I think the architecture diagram is a good idea, but this patch is big enough already; we can do that in a follow-up step.

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
-  * pganalyze Enterprise Server is delivered as a container image compatible with Docker / podman
-  * When using podman, make sure to replace `docker` with `podman` in the steps below
-* Setup a PostgreSQL database that can be used for storing statistics data
-  * PostgreSQL 11 or newer is required
-  * 50 GB of dedicated disk space or more
-* Optional: If you use the RDS integration, set up an appropriate IAM role that can be used to download logs and system metrics
-  * You can also assign the required policy to an existing EC2 instance and its instance role
-  * You will need the [collector IAM policy](/docs/install/amazon_rds/iam_policy)
+- **pganalyze Enterprise Server** — the central application that stores metrics, processes
+  statistics, and serves the web UI. It runs as a container (Docker, podman, or Kubernetes)
```
lfittl:
```suggestion
  statistics, and serves the web UI. It runs as one or more containers (Docker, podman, or Kubernetes)
```

enterprise/setup/index.mdx
lfittl: Yeah, I think the other spot is likely a bit outdated (and I assume comes from the previous guide), so maybe we should adjust it there to match the wording here?

I think we've seen more people run the collector separately in recent times (and it does work better for larger deployments, e.g. when multiple containers might also be used for the Enterprise Server install), so maybe useful to emphasize it a bit more.

The other thing to note is that the "works best for Amazon RDS / Aurora" isn't strictly true anymore, since we do also support passing the necessary variables for other platforms now, I believe.

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
-![](create_organization.png)
+**3. Configure network access**
 
-If authentication does not work, or you see an error message, please check the container's logs using `docker logs`.
+The following network connections need to be open between components:
 
-Choose an organization name of your choice (typically your company name), and then click "Add Server" on the next screen.
+| Connection | Port | Protocol |
+|---|---|---|
+| Monitored PostgreSQL databases &rarr; collector | 5432 | TCP |
```
lfittl, line 88: Shouldn't this be `collector -> Monitored PostgreSQL databases`? (since the collector initiates the connection)

enterprise/setup/docker.mdx
ryanbooz: It's actually funny you asking about this now that I read it. I do think we should assume at least 2 cores and 8GB. More and more of our enterprise PoCs... and certainly production environments... are processing larger queries and have higher throughput. Why introduce a resource limitation from the beginning when the cost difference isn't that big, right?

Thoughts?

enterprise/setup/index.mdx
ryanbooz: Good catch. I personally like having customers use separate collectors just because of the update cycle difference, but I know that's not historically been true. Anyone have a strong preference of which we promote most?

Honestly, I'd be more OK with the integrated collector if it were easier to get access/edit the config file in the UI. The forms are slow to populate, especially if you are adding multiple servers that are all similar (read replicas for instance). I've had multiple customers ask if they can duplicate and change or edit the file directly.

Oops. I just crossed into feature request land. My bad. 🙂

enterprise/setup/index.mdx
ryanbooz: This is still true. As far as I know, only RDS/Aurora has it included in `shared_preload_libraries` by default.

enterprise/upgrade/index.mdx
ryanbooz: No, I don't believe so. But agree, good for a follow-up after we verify.

enterprise/upgrade/index.mdx
ryanbooz: 💥 - I love it. You're absolutely right!

enterprise/setup/index.mdx
ryanbooz: Doh! Yep. I actually remember noticing that at one point and totally forgot to come back. Thanks!

enterprise/setup/docker.mdx
ryanbooz: @lfittl - do you have any preference here? Should we just say minimum 2 cores and 8GB?

enterprise/setup/index.mdx
ryanbooz:
```suggestion
| collector &rarr; Monitored PostgreSQL databases | 5432 | TCP |
```

enterprise/setup/index.mdx
ryanbooz: Yeah. I see what you mean. Good call.

enterprise/setup/index.mdx
ryanbooz: Didn't see that Lukas has replied. I agree and will get something committed for review.

enterprise/setup/docker.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+To learn more about adding additional team members, see [Account Management](/docs/accounts).
+
+We also recommend changing both the email and password of the admin user initially created (you can do so by clicking on "Admin" in the lower left of the screen).
+
+Additionally, you can review [all configuration settings](/docs/enterprise/settings) for the pganalyze Enterprise Server container.
+
+## Appendix: How to apply config changes
+
+When you make changes to the pganalyze container configuration in `/etc/pganalye.env`, make sure to fully restart the container afterwards, like this:
```
keiko713, line 204:
```suggestion
When you make changes to the pganalyze container configuration in `/etc/pganalyze.env`, make sure to fully restart the container afterwards, like this:
```

enterprise/upgrade/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+notes that apply across all deployment environments. For environment-specific steps, see
+the guides linked below.
+
+<ToC items={props.toc} />
+
+## General upgrade process
+
+Regardless of how you deploy the pganalyze Enterprise Server, every upgrade follows the
+same sequence:
+0. ** Review release notes** — check the [release notes](https://pganalyze.com/docs/enterprise/releases) for the new version (and any intermediate versions you are upgrading past). Releases with extra migration steps are noted in the following section, but release notes will help you know what to expect.
```
keiko713, line 21:
```suggestion
0. **Review release notes** — check the [release notes](https://pganalyze.com/docs/enterprise/releases) for the new version (and any intermediate versions you are upgrading past). Releases with extra migration steps are noted in the following section, but release notes will help you know what to expect.
```

Note: this works, but the rendering will likely start with 1. I think it's okay, but just wanted to note in case you really want to start with 0.

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
+over time.
 
-### Step 7: Add your first database to pganalyze
+#### Monitoring layers
 
-You can now either:
+pganalyze monitoring is made up of three layers. Enabling all three layers provides the most
+robust monitoring and optimization experience in pganalyze and is what we recommend. However, 
+some services (eg. Azure Event Hubs for streaming Postgres logs in Azure) may require assistance
+from other teams in your organization and cannot be setup initially. Each of these can layers can
```
keiko713, line 119:
```suggestion
from other teams in your organization and cannot be setup initially. Each of these layers can
```

enterprise/setup/index.mdx
```diff
@@ -1,210 +1,178 @@
...
 
-![](login.png)
+The pganalyze Enterprise Server is delivered as a container image. Choose the deployment
+method that fits your infrastructure. Each guide walks through pulling the image, configuring environment variables, initializing
+the database, creating an admin user, and starting the server. Resource requirements
+will vary — a proof-of-concept with a handful of servers needs far fewer resources than a
+production deployment monitoring dozens of instances.
 
-Now you can log in to the pganalyze interface using the generated credentials you've seen earlier when setting up the database. If successful you should see this screen:
+- [Docker on a VM](/docs/enterprise/setup/docker_vm) — the simplest option for evaluations
```
keiko713, line 76:
```suggestion
- [Docker on a VM](/docs/enterprise/setup/docker) — the simplest option for evaluations
```

enterprise/architecture.mdx
```diff
@@ -12,6 +12,6 @@ You can see two different options for the monitored environment below:
 
 **Option 2:** Self-managed Postgres servers running on-premise, in a virtual machine in the cloud, or using containerized deployments (e.g. with [CNPG](https://cloudnative-pg.io/)).
 
-This diagram is focused on larger setups that [run the collector separately](/docs/enterprise/setup/separate-collector-install), sending data to an internal, central installation of pganalyze Enterprise Server. For small deployments and evaluations we recommend the [simplified installation](/docs/enterprise/setup) (collector runs in the combined image and connects out).
+This diagram is focused on larger setups that [run the collector separately](/docs/enterprise/setup/separate-collector-install), sending data to an internal, central installation of pganalyze Enterprise Server. For small deployments and evaluations we recommend the [simplified installation](/docs/enterprise/setup/docker_vm) (collector runs in the combined image and connects out).
```
keiko713, line 15:
```suggestion
This diagram is focused on larger setups that [run the collector separately](/docs/enterprise/setup/separate-collector-install), sending data to an internal, central installation of pganalyze Enterprise Server. For small deployments and evaluations we recommend the [simplified installation](/docs/enterprise/setup/docker) (collector runs in the combined image and connects out).
```

keiko713 (approved):
Looks good! +1 on Maciek re: diagram, I think we can punt for now.

enterprise/setup/docker.mdx
ryanbooz: I'm going to leave for now. We can re-evaluate this and some code reuse stuff that is now shared between multiple pages once the site migration happens IMO.

enterprise/setup/index.mdx
msakrejda: Sorry for the delayed follow up: right, but maybe we should word this to avoid implying a restart is _always_ necessary. In many production situations, a database restart is a show-stopper for a trial.

PR #397, opened 2026-04-25
Update docs with cluster context

bisonhubert:
Implements https://github.com/pganalyze/pganalyze/issues/5703

### DIFF
--- a/accounts/permissions.mdx
+++ b/accounts/permissions.mdx
@@ -10,13 +10,18 @@ Every member of your organization gets to see the list of all other members in t
 but can't access any database information without having the **View** permission on them.
 
 Permissions are additive-only - that means if you grant a role that gives
-unrestricted access to all databases and servers, another restrictive permission
-that is also assigned will not restrict access for them.
+unrestricted access to all databases, servers, and server groups, another
+restrictive permission that is also assigned will not restrict access for
+them.
 
 ## View
 
 Allows the member to see statistics information for a database, including full
 query details as well as complete log information when using Log Insights.
+
+You can restrict the view permission to a specific database, server, or
+server group. A permission granted on a server group applies to every server and 
+database in the group.
 
 ## Modify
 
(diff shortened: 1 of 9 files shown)


query-performance/index.mdx
```diff
@@ -23,6 +23,18 @@ The list is sorted by the percentage of all runtime by default. You can search
 this list by part of the query text, or the name of the Postgres roles executing
 the query.
 
+### Cluster-wide query performance
+
+For servers grouped into a [cluster server
+group](/docs/accounts/server_groups), the same view is available across the
+whole cluster at `/servergroups/:id/queries`. The default page summarizes
```
keiko713: Do we have this pattern of sharing the path like this?

accounts/server_groups.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: 'Server Groups'
+backlink_href: /docs/accounts
+backlink_title: 'Account Management'
+---
+
+A server group is a way to group related servers within an organization, so
+that pganalyze can present them together and aggregate data across their
+members.
```
keiko713, line 9: Currently, we only support (have) the cluster server groups, and don't have the manual server groups, correct? It might be good to mention it someway.

accounts/permissions.mdx
```diff
@@ -10,33 +10,40 @@ Every member of your organization gets to see the list of all other members in t
 but can't access any database information without having the **View** permission on them.
 
 Permissions are additive-only - that means if you grant a role that gives
-unrestricted access to all databases and servers, another restrictive permission
-that is also assigned will not restrict access for them.
+unrestricted access to all databases, servers, and server groups, another
```
keiko713: Might be good to add a link to the server groups page from here.

keiko713 (approved):
A few comments but looking good!

query-performance/index.mdx
bisonhubert: Good catch. We don't reference URL paths elsewhere in the docs. I rewrote it to describe the navigation instead of the path.

accounts/server_groups.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+Today, pganalyze only supports cluster server groups, which are created
+automatically based on the collector's cluster ID.
```
msakrejda: It's not really the _collector_'s cluster id, so maybe something like
```suggestion
Today, pganalyze only supports cluster server groups, which are created
automatically based on the cluster ID determined by the collector (see below).
```

query-performance/index.mdx
```diff
@@ -23,6 +23,18 @@ The list is sorted by the percentage of all runtime by default. You can search
...
+
+For servers grouped into a [cluster server
+group](/docs/accounts/server_groups), the same view is available across the
+whole cluster. The default page summarizes query activity per database group 
+(databases sharing a name across cluster members). Pick a database group from 
+the table or Database picker to see top queries for that group, aggregated 
+across every cluster member that hosts that database. The drill-down offers the 
+same latency graph, statement-type filters, and time filter as the per-server
+view, and the query list adds a column showing which server each query came
+from.
```
msakrejda, line 36: Nice, thanks for adding this.

msakrejda (approved):

PR #399, opened 2026-05-06
Add docs for Disk Spill Due To Low work_mem

keiko713:
For https://github.com/pganalyze/pganalyze/pull/5785

PR #400, opened 2026-05-18
Use title case in section headers and fix directory churn

bisonhubert:
This started as a simple heading update for the recently added server groups page in account management. Headings should be in title case.

While testing that against pganalyze-web in https://github.com/pganalyze/pganalyze-web/issues/782  I noticed  `docs/directory.json` was repeatedly going dirty with a generated entry for the disk-spill insight added in #399. This commit adds the missing entry to stop the churn when Gatsby regenerates on each run.


accounts/server_groups.mdx
```diff
@@ -11,7 +11,7 @@ members.
 Today, pganalyze only supports cluster server groups, which are created
 automatically based on the cluster ID determined by the collector (see below).
 
-## Cluster server groups
+## Cluster-Based Server Groups
```
bisonhubert, line 14: > Cluster-based server groups are available ...

Ref https://pganalyze.com/blog/server-groups-and-cluster-wide-query-performance#try-it-now

lfittl (approved):

keiko713:
> I noticed docs/directory.json was repeatedly going dirty

sorry and thanks for the fix!

PR #405, opened 2026-06-25
Add Inefficient Index Use doc

keiko713:
Pairing PR of https://github.com/pganalyze/pganalyze/pull/5787


lfittl (approved):
We could probably refine that example a bit and provide more guidance on how to fix it beyond hint plan (often times the cause can be bloated indexes, and changing indexes also sometimes is the right choice), but for now I'll get this released as-is.

PR #409, opened 2026-07-07
Add missing Enterprise config for email, session cookies, and SSL settings

ryanbooz:


enterprise/settings.mdx
```diff
@@ -29,6 +30,60 @@ MAILER_URL=smtp://myusername:mypassword@example.com:25
 MAILER_FROM=pganalyze@example.com
 ```
 
+**Session cookies with multiple web pods**
```
lfittl: I wonder if we should consistently use "containers" instead of "pods", since that's the more general term - even though people are unlikely to run multiple containers outside of a Kubernetes environment.

enterprise/settings.mdx
```diff
@@ -29,6 +30,60 @@ MAILER_URL=smtp://myusername:mypassword@example.com:25
...
+An example SMTP configuration looks like this:
+
+```ini
+MAILER_URL=smtp://myusername:mypassword@example.com:25
+MAILER_SSL=none
+MAILER_SSL_VERIFY_MODE=none
+MAILER_AUTHENTICATION=login
+MAILER_DOMAIN=example.com
+MAILER_FROM=pganalyze@example.com
+```
```
lfittl, line 59: It might be good to also document how to send with an internal Microsoft Outlook instance set up for SMTP. I have notes that say the following works:

```
MAILER_SSL_VERIFY_MODE=none
MAILER_AUTHENTICATION=login

username without part after the @ (only name in front of @)
```

enterprise/settings.mdx
```diff
@@ -29,6 +30,60 @@ MAILER_URL=smtp://myusername:mypassword@example.com:25
...
+
+## SSL / TLS configuration
+
+The `web` container runs an NGINX reverse proxy that offers SSL support on port `5001`. By default a self-signed certificate is auto-generated, but you can provide your own certificate and key using the following settings:
+
+* `SSL_CERT`: The full SSL certificate, passed in directly (optional)
+* `SSL_KEY`: The full SSL private key, passed in directly (optional)
+* `SSL_CERT_PATH`: Path to a custom SSL certificate inside the container, e.g. a volume mount (optional, requires [2025.03.0](/docs/enterprise/releases/2025-03-0) or newer)
+* `SSL_KEY_PATH`: Path to a custom SSL private key inside the container, e.g. a volume mount (optional, requires [2025.03.0](/docs/enterprise/releases/2025-03-0) or newer)
```
lfittl: I wonder if we really need the version references - they are a nice touch, but the release is more than a year old at this point, and usually this kind of setting is configured when doing a new installs.

lfittl (approved):

enterprise/settings.mdx
ryanbooz: Oh, good call. I've also debated making this bold vs. a H2 or H3. Either of those ends up in the ToC at the top which didn't feel like it was worth that. H3 gets indented as a child, and that also seemed odd.

OK as bold?

enterprise/settings.mdx
ryanbooz: Fair point. I still saw 2025 and thought... "maybe". But you're right I think.

PR #412, opened 2026-07-10
Use the isolated hljs instance for docs CodeBlock

keiko713:
With Astro migration, we started doing `{ classPrefix: 'token ' }` to hljs, but because this was done at the global level, this was also affecting the app's SQL highlighting behavior, causing not to highlight in app.
With this change, it's gonna use the isolated hljs instance so the app side will not be affected.


msakrejda (approved):
Ah, good catch.

msakrejda:
I'll merge this to get the fix in.

msakrejda:
Confirmed that this fixes the issue in the app (it looks like highlighting is only broken in some places) and does not break highlighting in the public site.

PR #413, opened 2026-07-16
Update docs for alert destination changes

msakrejda:
- Add missing Advisor Insight docs
- Update docs for Alert Destination changes


components/CheckDocumentation/queries/AdvisorInsight.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const AdvisorInsightTrigger: React.FunctionComponent<CheckTriggerProps> = () => {
+  return (
+    <p>
+      Detects queries whose execution plans from the last seven days contain a
+      supported Query Advisor insight and creates an issue with severity "info".
+      Currently detected problems are inefficient nested loops, wrong
+      indexes due to <code>ORDER BY</code>, inefficient index use, late join
+      filters with <code>OR</code>, or disk spills due to low{" "}
+      <code>work_mem</code>.
```
keiko713, line 17: I'm slightly worried this gets outdated (also I don't think we particularly make them using `<code>` as a insight name, but it's okay either way for me).

checks/alert-destination-editor.png
keiko713, line 1: I can't stop wondering why Maciek's user name below there is "Av Macieck Sak..." 🤔 (I think nobody cares so this screenshot is good as is)

keiko713 (approved):
Looking good!

components/CheckDocumentation/queries/AdvisorInsight.tsx
msakrejda: It's not really an insight name here: it's referring to the Postgres setting. At least that's how I was thinking about it.

checks/alert-destination-editor.png
msakrejda: Oh ugh. Let me see if I can fix that.

keiko713: Ohhhh it was avatar's "Av"!

PR #417, opened 2026-07-28
Update ActiveQuery doc for new ignore options

keiko713:


components/CheckDocumentation/connections/ActiveQuery.tsx
```diff
@@ -25,10 +29,26 @@ const ActiveQueryTrigger: React.FunctionComponent<CheckTriggerProps> = ({
...
+          configuration.
+        </p>
+      )}
+      {ignoreMaintenance && (
+        <p>
+          Also ignores non-blocking maintenance commands (<code>VACUUM</code>,
+          excluding <code>VACUUM FULL</code>, and <code>ANALYZE</code>), based
+          on this check's configuration.
+        </p>
+      )}
```
msakrejda: These default to off in `util/checks/index.ts` so they'll never show up in the public docs. Maybe we should also pass a `isActualConfig` or something to our check docs so they can distinguish actual config from default config? Right now they have to render in two places with no real indication which is which.

msakrejda (approved):
I've realized the docs have a wart that means the new info won't render at all, but it's not worth yak shaving into here. I can take the follow-up task of giving the docs components more context unless you want it.

components/CheckDocumentation/connections/ActiveQuery.tsx
keiko713: Uhmm good point. Looking at other docs, actually we don't have this pattern (of not displaying the thing if the config is off), so I tweaked a bit. How does this look?

Image

msakrejda: Yeah, I think that reads okay and is probably clearer. I still think calling out default config versus actual config is worthwhile, but separately.

keiko713: "`1800` seconds" is also a default config, so we can think about showing what is default separately indeed.

PR #419, opened 2026-08-11
New documentation for setting up Neon and Supabase as supported system types in collector

ryanbooz:
Neon sections added for Install and Log Insights

Supabase sections added for Install, Log Insights, and EXPLAIN

Includes some related visual fixes that were pre-existing bugs


explain/setup/supabase/01_auto_explain_check.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+The recommended Automated EXPLAIN configuration for Supabase is to use the Postgres
+`auto_explain` contrib module. This ensures accurate and detailed plan information,
+including runtime information (as gathered by EXPLAIN ANALYZE). With our recommended
+configuration (presented in a later step), performance overhead is minimal.
+
+Unlike most other platforms, Supabase does not let you change `shared_preload_libraries`
+or use `ALTER SYSTEM` yourself — instead, most Postgres configuration on Supabase
+(including all `auto_explain.*` settings) is set at the **role level**, using
+`ALTER ROLE ... SET`. We'll cover that in Step 3.
```
msakrejda:
```suggestion
```
This doesn't really say anything useful.

explain/setup/supabase/_02_actually_enable.mdx
```diff
@@ -0,0 +1,6 @@
+On Supabase, `shared_preload_libraries` is managed by the platform, and projects are not
+able to run `ALTER SYSTEM` themselves. If `auto_explain` doesn't show up in
+`shared_preload_libraries` for your project, [contact Supabase support](https://supabase.com/dashboard/support/new)
+and ask them to add `auto_explain` to your project's preload libraries.
```
msakrejda:
```suggestion
On Supabase, `shared_preload_libraries` is not user-configurable, but it includes `auto_explain`
by default. If `auto_explain` doesn't show up in `shared_preload_libraries` for your project, contact Supabase support.
```

This has been included by default for [over four years now](https://github.com/supabase/postgres/pull/258), so I think we can streamline the instructions and assume the happy path.

explain/setup/supabase/03_review_settings.mdx
```diff
@@ -0,0 +1,43 @@
...
+The right configuration will depend on your server and workload,
+but we've found the following is a good starting point:
+
+## Apply these settings at the role level
+
+Because Supabase doesn't allow `ALTER SYSTEM`, these settings need to be applied per role,
+using `ALTER ROLE ... SET`, for **every role your application connects as** (the `postgres`
+role alone is not enough — queries only get an EXPLAIN plan logged if the connecting role
+has these settings applied). Run the following as the `postgres` user, once for each
+application role (replace `your_app_role` with the actual role name):
```
msakrejda: It doesn't make sense to have this section after the
> ...the following is a good starting point:

That's supposed to introduce the settings themselves. How about reorganizing this into

```suggestion
There are a number of `auto_explain` settings that allow you to configure which queries
will generate EXPLAIN plans. You can read the full Postgres documentation about
these <AutoExplainLink version={props.pgMajorVersion}>here</AutoExplainLink>.

## Apply these settings at the role level

Because Supabase doesn't allow `ALTER SYSTEM`, these settings need to be applied per role,
using `ALTER ROLE ... SET`, for **every role your application connects as** (the `postgres`
role alone is not enough — queries only get an EXPLAIN plan logged if the connecting role
has these settings applied). Run the following as the `postgres` user, once for each
application role (replace `your_app_role` with the actual role name).

The right configuration will depend on your server and workload,
but we've found the following is a good starting point:
```

? Or alternately, I think we can skip the "apply at role level" subsection heading altogether and just include that in the body:


```suggestion
There are a number of `auto_explain` settings that allow you to configure which queries
will generate EXPLAIN plans. You can read the full Postgres documentation about
these <AutoExplainLink version={props.pgMajorVersion}>here</AutoExplainLink>.

Because Supabase doesn't allow `ALTER SYSTEM`, these settings need to be applied per role,
using `ALTER ROLE ... SET`, for **every role your application connects as** (the `postgres`
role alone is not enough — queries only get an EXPLAIN plan logged if the connecting role
has these settings applied). Run the following as the `postgres` user, once for each
application role (replace `your_app_role` with the actual role name).

The right configuration will depend on your server and workload, but we've found the
following is a good starting point:
```

install/neon/02_create_monitoring_user.mdx
```diff
@@ -0,0 +1,34 @@
...
+backlink_title: 'Installation Guide'
+---
+
+import MonitoringUserSetupInstructions from "../../components/MonitoringUserSetupInstructions";
+
+We recommend creating a separate monitoring user, rather than using the default role
+that was created for you when the project (or branch) was created.
+
+Each branch starts out with a single database, named `neondb` by default (or a custom
+name if you chose one when creating the project). Connect using your
```
msakrejda: Maybe
```suggestion
Each branch starts out with a single database (`neondb` by default). Connect using your 
```
? The "by default" already implies that there's a non-default thing you can do, and users can probably figure out what database they want to monitor.

install/neon/03_install_the_collector_docker.mdx
```diff
@@ -0,0 +1,18 @@
...
+backlink_href: /docs/install
+backlink_title: 'Installation Guide'
+---
+
+## Downloading the collector container
+
+Pull the Docker image like this:
+
+```bash
+docker pull quay.io/pganalyze/collector:stable
```
msakrejda, line 13: This just downloads the image. This is not equivalent to actually installing it like with our package install. 

Although reviewing the, e.g., Azure and AWS instructions we kind of have the same problem there, so maybe we can go with this for now and improve these all in a separate sweep as a follow-up step.

install/supabase/02_create_monitoring_user.mdx
```diff
@@ -0,0 +1,38 @@
...
+<MonitoringUserSetupInstructions adminUsername='postgres' databaseName='postgres' />
+
+Supabase installs extensions like `pg_stat_statements` into a dedicated `extensions`
+schema, rather than `public`. Since the `pganalyze` user doesn't have access to that
+schema by default, grant it explicitly (still connected as **postgres**):
+
+```sql
+GRANT USAGE ON SCHEMA extensions TO pganalyze;
+GRANT SELECT ON extensions.pg_stat_statements TO pganalyze;
+```
```
msakrejda, line 27: We don't need to put `extensions` in the pganalyze user's `search_path`, right?

install/supabase/03_install_the_collector_docker.mdx
```diff
@@ -0,0 +1,18 @@
...
+backlink_title: 'Installation Guide'
+---
+
+## Downloading the collector container
+
+Pull the Docker image like this:
+
+```bash
+docker pull quay.io/pganalyze/collector:stable
+```
```
msakrejda, line 14: Same comment as on the neon version of this, although like above, we can defer this to a follow-up.

install/supabase/04_configure_the_collector_docker.mdx
```diff
@@ -0,0 +1,73 @@
...
+   `aws-0-<region>.pooler.supabase.com`. You can find it under **Project Settings &rarr; Database**
+   in the Supabase dashboard (choose the "Session pooler" tab)
+3. The `DB_PORT` is `5432` for the **session pooler** — do not use port `6543`, which is the
+   transaction pooler and doesn't support the session-level features the collector relies on
+4. The `DB_NAME` is the database you want to monitor (`postgres` by default)
+5. The `DB_USERNAME` and `DB_PASSWORD` should be the credentials of the monitoring user we
+   created in Step 1. Note the pooler requires the `<role>.<project-ref>` username format,
+   so this is `pganalyze.your-project-ref`, not just `pganalyze`
+
+If you're paying for Supabase's IPv4 add-on and prefer to connect directly instead, use the
```
msakrejda, line 58:
```suggestion
If you're using for Supabase's IPv4 add-on and prefer to connect directly instead, use the
```
Whether you're using it is more relevant here. Or we could say something like
```suggestion
If prefer to connect directly instead, you can leverage [Supabase's IPv4 add-on](https://supabase.com/docs/guides/platform/ipv4-address) and use the
```
to focus on your preferred setup rather than your current situation.

install/supabase/04_configure_the_collector_package.mdx
```diff
@@ -0,0 +1,75 @@
...
+   `aws-0-<region>.pooler.supabase.com`. You can find it under **Project Settings &rarr; Database**
+   in the Supabase dashboard (choose the "Session pooler" tab)
+3. The `db_port` is `5432` for the **session pooler** — do not use port `6543`, which is the
+   transaction pooler and doesn't support the session-level features the collector relies on
+4. The `db_name` is the database you want to monitor (`postgres` by default)
+5. The `db_username` and `db_password` should be the credentials of the monitoring user we
+   created in Step 1. Note the pooler requires the `<role>.<project-ref>` username format,
+   so this is `pganalyze.your-project-ref`, not just `pganalyze`
+
+If you're paying for Supabase's IPv4 add-on and prefer to connect directly instead, use the
```
msakrejda, line 60: Same as above.

install/supabase/_04_find_connection_info.mdx
```diff
@@ -0,0 +1,18 @@
+import imgSupabaseGetConnected from './supabase-get-connected.png'
+import imgSupabaseConnectionDetails from './supabase-connection-details.png'
+
+## Finding your Supabase connection information
+
+The collector keeps an ongoing connection and we generally recommend connecting it through
+the Supavisor **session pooler** (port `5432` on the pooler hostname), rather than the direct connection.
+Supabase's direct connection is IPv6-only by default, and reaching it over IPv4 requires paying for the
+[IPv4 add-on](https://supabase.com/docs/guides/platform/ipv4-address).
```
msakrejda: This implies that connecting over IPv6 doesn't work without actually claiming that outright. We should say either more or less about this.

log-insights/setup/neon/01_set_up_opentelemetry_logs.mdx
```diff
@@ -0,0 +1,79 @@
...
+db_name = your_database_name
+...
+db_log_otel_server = 0.0.0.0:4318
+```
+
+Neon's infrastructure needs to reach this endpoint over the public internet, so keep the following in mind:
+
+* Run the collector somewhere with a stable, publicly reachable address (a small VM works well)
+* Although not required, we highly recommend running the collector behind a TLS-terminating reverse
+  proxy or load balancer in front of the `db_log_otel_server` port to provide a `https://` endpoint URL
```
msakrejda:
```suggestion
* We highly recommend running the collector behind a TLS-terminating reverse
  proxy or load balancer in front of the `db_log_otel_server` port to provide a `https://` endpoint URL
```
We should really discourage insecure setups like that.

log-insights/setup/supabase/01_set_up_log_drain.mdx
```diff
@@ -0,0 +1,84 @@
...
+backlink_title: 'Log Insights: Setup'
+---
+
+import imgSupabaseLogDrain from './supabase-create-log-drain.png'
+import imgSupabaseOtlpSettings from './supabase-otlp-settings.png'
+
+Supabase allows you to configure a [Log Drain](https://supabase.com/docs/guides/telemetry/log-drains) that
+pushes logs to an OpenTelemetry (OTLP) HTTP endpoint that the pganalyze collector exposes.
+
+**Log Drains require the Supabase Pro plan or higher** (not available on the Free plan).
```
msakrejda:
```suggestion
**Log Drains require the Supabase Pro plan or higher**.
```

log-insights/setup/supabase/01_set_up_log_drain.mdx
```diff
@@ -0,0 +1,84 @@
...
+db_port = 5432
+db_username = pganalyze.your-project-ref
+...
+db_log_otel_server = 0.0.0.0:4318
+```
+
+Supabase's infrastructure needs to reach this endpoint over the public internet, so keep the following in mind:
+
+* Run the collector somewhere with a stable, publicly reachable address (a small VM works well)
+* Although not required, we highly recommend running the collector behind a TLS-terminating reverse
```
msakrejda:
```suggestion
* We highly recommend running the collector behind a TLS-terminating reverse 
```
as above.

msakrejda (commented):
This is great, thanks for working on it. I left a few comments, but overall it looks good.

install/supabase/_04_find_connection_info.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+You can find your session pooler
+connection information by clicking on the **Get Connected** button in the Supabase dashboard:
+
+<img src={imgSupabaseGetConnected} alt="Supabase dashboard Get Connected button" className="img-polaroid" style={{ width: '520px' }} />
+<br/><br/>
+Then select the **Session pooler** option under the **Direct Connection String** tab.
+
+<img src={imgSupabaseConnectionDetails} alt="Supabase connection details showing the Session pooler option" className="img-polaroid" style={{ width: '500px' }} />
```
keiko713, line 16: You wanna add `<br/><br/>` here too, as it's currently really tight.

Image

install/neon/02_create_monitoring_user.mdx
```diff
@@ -0,0 +1,34 @@
...
+---
+
+import MonitoringUserSetupInstructions from "../../components/MonitoringUserSetupInstructions";
+
+We recommend creating a separate monitoring user, rather than using the default role
+that was created for you when the project (or branch) was created.
+
+Each branch starts out with a single database, named `neondb` by default (or a custom
+name if you chose one when creating the project). Connect using your 
+project's default role which is a member of the `neon_superuser` role, and run the
```
keiko713: Is this different from the way you mentioned in the enabling pgss?

> using the connection string from the [Neon Console](https://console.neon.tech) (**Connect** button on the Project Dashboard)

install/neon/03_install_the_collector.mdx
```diff
@@ -0,0 +1,50 @@
...
+      ]}
+    />
+  )
+}
+
+The collector runs on a VM or container of your choosing, and connects to your
+Neon branch over the network. We recommend at least 1GB of RAM (you may need
+more to handle large schemas or complex query workloads).
+
+**Note:**: If you want to collect logs from your Neon branch ([configured separately](/docs/log-insights/setup/neon/01_set_up_opentelemetry_logs)), you must run the
```
keiko713, line 44: Nice to mention this here 👍

explain/setup/supabase/_02_actually_enable.mdx
keiko713: I am personally not sure about having the contact support link here. Given that this very less likely happens, I think saying "contact Supabase support" without a link is okay.
(I'm not a fan of having a link to the vendor's support page, and this is the only place doing so)

explain/setup/index.mdx
```diff
@@ -30,17 +30,26 @@ Log-based EXPLAIN is not guaranteed to show the same plan that was executed, and
...
+Heroku Postgres               | Yes               | Yes (Recommended)   |
+Crunchy Bridge                | Yes               | Yes (Recommended)   |
+Aiven                         | Yes               | No                  |
+Supabase                      | Yes               | Yes (Recommended)\* |
+Neon                          | No                | No                  |
+Self-managed VM               | Yes               | Yes (Recommended)   |
+Kubernetes                    | No                | No                  |
+Other PaaS                    | Contact support   | Contact support     |
+
+\* Supabase only supports configuring `auto_explain` settings at the role level (rather than
```
keiko713, line 47: I won't look good so no need to tweak here, but there is a markdown native footnote you could add, like `[^1]`, FYI.

log-insights/setup/neon/neon-otlp-settings.png
keiko713, line 1: Neon does support https, right? The screenshot's example URL is `http://`, which makes me wonder if it does. Well, but maybe updating the screenshot is too much work, so no need to tweak.

keiko713 (approved):
+1 on Maciek's review too, and I also added some nits. Approving, assuming you'll address them.

explain/setup/supabase/01_auto_explain_check.mdx
ryanbooz: Fair enough.

explain/setup/supabase/03_review_settings.mdx
ryanbooz: Agreed

explain/setup/supabase/_02_actually_enable.mdx
ryanbooz: I think you're right about the link.

Part of keeping this step was for consistency initially and then I figured we could (eventually) do a review of the overall flow for each platform. I agree that there's really no reason `auto_explain` shouldn't be available and it's not clear how it would be unsupported in a particular account.

explain/setup/index.mdx
ryanbooz: 👍🏻

install/neon/03_install_the_collector_docker.mdx
ryanbooz: Yeah. I started from what we have and figured I'd just keep it consistent for now. I think an overall revamp is definitely in order soon.

install/supabase/02_create_monitoring_user.mdx
ryanbooz: Nope. The collector figure it out.

install/supabase/_04_find_connection_info.mdx
ryanbooz, line 16:
```suggestion
<img src={imgSupabaseConnectionDetails} alt="Supabase connection details showing the Session pooler option" className="img-polaroid" style={{ width: '500px' }} />
<br/><br/>
```

log-insights/setup/neon/neon-otlp-settings.png
ryanbooz: Updated.

install/neon/02_create_monitoring_user.mdx
```diff
@@ -0,0 +1,33 @@
...
+backlink_title: 'Installation Guide'
+---
+
+import MonitoringUserSetupInstructions from "../../components/MonitoringUserSetupInstructions";
+
+We recommend creating a separate monitoring user, rather than using the default role
+that was created for you when the project (or branch) was created.
+
+Each branch starts out with a single database (`neondb` by default). Connect using your 
+project's default role which is a member of the `neon_superuser` role, and run the
```
ryanbooz:
```suggestion
Each branch starts out with a single database (`neondb` by default). Connect using the connection string from the Neon Console (Connect button on the Project Dashboard)
```

install/supabase/_04_find_connection_info.mdx
```diff
@@ -0,0 +1,19 @@
+import imgSupabaseGetConnected from './supabase-get-connected.png'
+import imgSupabaseConnectionDetails from './supabase-connection-details.png'
+
+## Finding your Supabase connection information
+
+The collector keeps an ongoing connection and we generally recommend connecting it through
+the Supavisor **session pooler** (port `5432` on the pooler hostname), rather than the direct connection.
+Supabase's direct connection is IPv6-only by default, and reaching it over IPv4 requires paying for the
```
ryanbooz:
```suggestion
The collector currently uses IPv4 to connect to your database, but by default, Supabase's direct connection is IPv6-only. Therefore, we generally recommend connecting through the Supavisor **session pooler** (port `5432` on the pooler hostname), rather than the direct connection unless you are already paying for the IPv4 direct connection add-on in your Supabase project. 
```

PR #420, opened 2026-08-13
Enterprise Server release 2026.08.0

lfittl:


enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+backlink_href: /docs/enterprise/releases
+backlink_title: 'pganalyze Enterprise Server: Release Changelog'
+---
+
+**Docker tag:** `quay.io/pganalyze/enterprise:v2026.08.0`
+
+## Notes
+
+- This release requires a database migration when coming from older releases, be sure to follow the [recommended upgrade steps](/docs/enterprise/upgrade)
+- **Important:** If you are currently running v2023.08.0 or older, please [reach out to us](/contact) to walk you through the upgrade process.
```
msakrejda, line 12: We should have some internal artifact documenting how to walk users through for the sake of those on support.

enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+      recommendations. These queries are now analyzed correctly and included in Index Advisor.
+  - Support uppercase column names in extended statistics {/* 5991 */}
+  - Prefer role-defined `search_path` if present when resolving schema objects {/* 6067 */}
+  - Unused index check: Raise maximum look back period from 35 days to 400 days {/* 5984 */}
+    - This allows configuring the unused index threshold to be significantly higher, to avoid incorrectly flagging
+      indexes that are used e.g. once a year. The default remains at 35 days.
+* EXPLAIN plans and Automated EXPLAIN:
+  - Support fractional rows and Index Searches in plan output {/* 6007 */}
+  - Fix plan text parser for Aurora Adaptive joins {/* 5895 */}
+  - Do nothing when duplicate query plans are detected {/* 5946 */}
```
msakrejda:
```suggestion
  - Avoid storing duplicate query plans {/* 5946 */}
```

enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+  - Only count up to 10k plans for plan counts, to avoid slow queries {/* 5943 */}
+  - Fix error when viewing large EXPLAIN plans {/* 6054 */}
+  - Support deeply nested EXPLAIN plans {/* 6078 */}
```
msakrejda: Maybe we should combine these two? Reviewing the source changes, I don't think it's useful to have two separate entries for the same release.

```suggestion
  - Fix error when viewing large EXPLAIN plans {/* 6054, 6078 */}
```

msakrejda: And should this be under bug fixes below? Although... were these both only fixes to issues introduced by the Rails upgrade? That's only in this release, so maybe the entries are moot and we should drop them?

enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+    - Previously, searching the query list for a term with no matches showed an unrelated setup hint suggesting that
+      query statistics were not being collected.
+  - Guard against all statement types being turned off {/* 5941 */}
+    - Previously, unchecking all three statement type filters ("SELECT", "INSERT, UPDATE, DELETE", and "DDL &amp; other")
+      hid all query performance data and showed an unrelated setup hint instead. Since the selection is persisted in
+      browser LocalStorage, this would also persist across visits.
+    - The change guards against all statement types being turned off, and automatically turns on SELECT statements for
+      users who may have previously turned off all statement types.
+* Graph:
+  - Avoid showing data points that are too far from the cursor in graph tooltips {/* 5906 */}
```
msakrejda: This is probably my wording, but it might be clearer as 
```suggestion
  - Skip series in graph tooltips when their nearest data points are too far from the cursor {/* 5906 */}
```

enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+  - Handle undefined interaction points on hover {/* 5907 */}
```
msakrejda: I think this was just avoiding console errors on hover; do we want to mention it? The original intentionally omits a CHANGELOG entry per our style guide ("When in doubt, omit the entry"). It's technically user-visible, so no objection if you want to keep it.

enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+* Render longer database names in database menu {/* 6041 */}
+* Log Insights
+  - Fix log line ingest byte size issue {/* 6114 */}
+    - If the collector sent a log snapshot with an incorrect byte size for a log file, the ingest code could drop log
+      lines beyond the reported "end" of the file. Log file sizes are now recomputed rather than trusted as sent.
+  - Explicitly split uploaded data by minute for packed format to support higher log download intervals {/* 6093 */}
```
msakrejda, line 131: Is this a fix? Reviewing the source change, I understood this was laying ground for an optimization.

enterprise/releases/2026-08-0.mdx
```diff
@@ -0,0 +1,180 @@
...
+* Fix table sorting on Connections page {/* 5981 */}
+* Improve SAML integration setup with clearer error messages {/* 5993 */}
+* Fix missed aggregate statistics (1 hour data points) due to bug in compaction process {/* 6122 */}
+* Reduce rate of re-analyzing of queries with previously failed/outdated query analysis {/* 6043 */}
```
msakrejda: This seems like a Performance or Notes item rather than a bugfix.

msakrejda (approved):
Left a bunch of minor comments, but looks good.

enterprise/releases/2026-08-0.mdx
lfittl: Not sure if its really needed - I think at this point almost everyone has upgraded - I could have already dropped this note I think (will do with the next release). The main reason for including it was because of the multiple query stats format changes (row-based to arrays to pco).

enterprise/releases/2026-08-0.mdx
lfittl: I think the first fix was independent of the Rails upgrade, but the second one was basically re-applying the fix because of the Rails upgrade. I'll take your suggestion to merge them.

enterprise/releases/2026-08-0.mdx
lfittl: Yeah, that's fair - I ended up pulling a lot more things from commit messages, but this one is good to omit.

enterprise/releases/2026-08-0.mdx
lfittl: Yeah, it's a fix for the higher log download intervals change in the collector - without it the data could be missed (since the data would be misplaced in the wrong day partition).

enterprise/releases/2026-08-0.mdx
lfittl: Yeah, that's fair, I'll move it to Performance.

PR #424, opened 2026-09-11
Add server identity page

keiko713:
This page will help users to understand how we recognize the server, which will be useful when they want to keep the historical metrics/stats when moving the servers.

Note that the current backend still _does not_ work the way it's documented here, that one needs to be merged before publishing this.

https://github.com/pganalyze/pganalyze/pull/6217


collector/server-identity.mdx
```diff
@@ -0,0 +1,226 @@
...
+Part of the identity changes when you replace an instance, move a database to a different hosting
+provider, or move it to a different AWS account. The collector then reports an identity that
+pganalyze has not seen before. By default this creates a *new* server, leaving the original one
+behind with all of the previous history. To avoid that, you can tell the collector what the
+previous identity was, using the `api_system_id_fallback`, `api_system_type_fallback`, and
+`api_system_scope_fallback` settings. pganalyze then finds the existing server and updates it to
+the new identity, so the history carries over.
+
+```toc
+```
```
seanlinsley, line 21: Should there be something here?

seanlinsley (approved):

collector/server-identity.mdx
keiko713: This will render the auto-generated TOC like below:

image

